PR #369, opened 2014-06-20
Kmem rework

ryao:
This refactoring is intended to simplify the code while making it easier to share code between ZoL and other Open ZFS platforms.


behlendorf:
Closing.  The kmem rework was merged some time ago.

PR #372, opened 2014-07-12
random_get_pseudo_bytes() need not provide cryptographic strength entropy

ryao:
Perf profiling of dd on a zvol revealed that my system spent 3.16% of
its time in random_get_pseudo_bytes(). No SPL consumers need
cryptographic strength entropy, so we can reduce our overhead by
changing the implementation to utilize a fast PRNG.

The Linux kernel did not export a suitable PRNG function until it
exported get_random_int() in Linux 3.10. While we could implement an
autotools check so that we use it when it is available or even try to
access the symbol on older kernels where it is not exported using the
fact that it is exported on newer ones as justification, we can instead
implement our own pseudo-random data generator. For this purpose, I have
written one based on a pseudo-random number generator proposed in a
paper by Sebastiano Vigna that itself was based on work by the late
George Marsaglia.

http://arxiv.org/pdf/1402.6246v2.pdf

Profiling the same benchmark with a variant of this patch that did not
disable interrupts showed that time spent in random_get_pseudo_bytes()
dropped to 0.06%. That is a factor of 50 improvement.

### DIFF
--- a/module/spl/spl-generic.c
+++ b/module/spl/spl-generic.c
@@ -41,6 +41,8 @@
 #include <sys/kstat.h>
 #include <sys/file.h>
 #include <linux/ctype.h>
+#include <sys/disp.h>
+#include <sys/random.h>
 #include <linux/kmod.h>
 #include <linux/math64_compat.h>
 #include <linux/proc_compat.h>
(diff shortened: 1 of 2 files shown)


sempervictus:
Any chance this could get updated for current revision? This still has a lot of the code cleaned up last fall attached. Thanks

ryao:
@sempervictus This was near the bottom of my priorities, but it is refreshed now.

ryao:
I have pushed a patch into the pull request to replace the 64-bit xorshift generator with a newer 128-bit xorshift+ generator that I found in literature:

http://vigna.di.unimi.it/ftp/papers/xorshiftplus.pdf

The numbers produced by it pass more stringent tests for statistical randomness than those that I existed when I familiarized myself with the topic of random number generation almost a decade ago. It is also very fast.

The following website claims that it takes 1.12 ns per 64bits on an Intel Core i7-4770 in a presumably single threaded benchmark. That is ~7.14GB/s per core.

http://xorshift.di.unimi.it/#speed

The design of `random_get_pseudo_bytes()` should prevent it from obtaining that level of speed in practice, but it should still be very fast. It should also outperform Linux's `get_random_bytes()` by one or two orders of magntiude.

module/spl/spl-generic.c
```diff
@@ -502,6 +609,44 @@ spl_kvmem_init(void)
...
+	int i;
+
+	get_random_bytes(s, sizeof (s));
+
+	if (s[0] == 0 && s[1] == 0) {
+		if (jiffies != 0) {
+			s[0] = jiffies;
+			s[1] = ~0 - jiffies;
+		} else {
+			(void) memcpy(s, "improbable seed", sizeof (s));
```
ryao: This is an incredibly improbable code path, but if we somehow hit it, it should be logged to dmesg as `0x006465657320656C6261626F72706D69`.

dweeezil:
@ryao  When I tried a simple "dd to a zvol" test as you mentioned in the original issue report, I noticed the hot call path (3.X% as well) was the ```vd == NULL``` case in ```vdev_mirror_map_alloc()```.  Was this the path you observed?  It seems this case could easily get away with (the equivalent of) ```gethrtime() % 3``` rather than relying on a proper PRNG.  Of course it's still a very Good Idea to have as efficient a PRNG as possible for, say, GUID generation etc.

ryao:
@dweeezil That is consistent with my recollection. Something like `gethrtime() % 3` could work, but it has some downsides versus a simple per-cpu PRNG:

1. Having a slow `random_get_pseudo_bytes()` would allow for future patches merged from illumos-gate to introduce regressions in other hot code paths that a patch today switching to `gethrtime() % 3` would not prevent, so we would just be giving ourselves more work to do. Reimplementing `random_get_pseudo_bytes()` with a per-CPU PRNG avoids that entirely.
2. `gethrtime()` uses memory barriers and maybe locks too depending on the clock source. Both are avoidable via a per-CPU PRNG, so replacing `random_get_pseudo_bytes()` with `gethrtime()` in hot code paths would just give a future person working on NUMA scalability more work to do.
3. The kernel code behind the `gethrtime() % 3` makes many function invocations, uses memory barriers and has different callpaths on different systems thanks to its ability to use different time sources through a pluggable function pointer. In comparison, a per-CPU PRNG puts only ~40 lines of code in `random_get_pseudo_bytes()`, eliminates recursion into kernel text and would have consistent behavior across all systems. That should minimize any amount of thinking we need to do about this in the future from the perspective of performance analysis.
4. Unless we are absolutely certain of the quality of `gethrtime() % 3` in tests for statistical randomness across all system hardware combinations and workloads, it would potential allow system-specific details to give rise to bias that could not be easily reproduced elsewhere. The PRNG is known to be free of bias, so this would not be a concern for it and consequently, minimizes the chance of needing to think about this again.

I probably have over-analyzed this, but it was fun and I like the idea of improving this to the extent that the topic of pseudo-random number generation should never need to be touched again.

ryao:
I merged the two into one commit.

behlendorf:
@ryao if you open a dummy PR against ZFS which includes the following line in the commit message we can get some testing on this.

```
Requires-spl: refs/pull/372/head
```

ryao:
@behlendorf I did a quick build test locally, which revealed that we had a C99ism, but this is otherwise fine. I have opened zfsonlinux/zfs#4321 with a dummy commit for the purpose of testing this.

dweeezil:
@ryao  All your points regarding ```gethrtime()``` are well taken.  Since my comments regarding the use of a PRNG in ```vdev_mirror_map_alloc() could be considered an issue hijack, I'll save any others for the OpenZFS list or a separate ZFS-specific issue.

ryao:
@dweeezil This was originally motivated by the excessive CPU time taken in `vdev_mirror_map_alloc()`. When I originally looked at this, I felt that modifying either `vdev_mirror_map_alloc()` or `spa_get_random(c)` would be unnecessary if `random_get_pseudo_bytes()` were updated.

ryao:
I made a small change to the patch so that "improbable seed" would be printed as 0x696D70726F6261626C65207365656400 on both big endian and little endian systems.

ryao:
This has passed the buildbot. I only had to kick it about 3 times. :/

behlendorf:
This LGTM, @dweeezil if you're also OK with this change I'll get it merged now that it's passing all the testing.

dweeezil:
LGTM, including the rationale explained in the commit comment.

PR #391, opened 2014-09-18
Add support for signing kernel modules

behlendorf:
Linux kernels which are built with CONFIG_MODULE_SIG support the
cryptographic signing of modules.  When in "permissive" mode
unsigned modules or signed modules with an unknown key will by
allowed to load but will mark the kernel as tainted.  When in
"restrictive" mode only modules signed with a known key will be
allowed to load.

This patch updates the build system so that it will automatically
sign modules if the required public and private keys are available.
These keys are generated as part of the kernel build process and
are named "signing_key.priv" and "signing_key.x509".  Placing a
copy of these keys in the ./modules/ directory will cause the
modules to be signed as part of the build process.  Alternately,
the SECKEY and PUBKEY variables may be set to specify a different
location.

make SECKEY=~/signing_key.priv PUBKEY=~/signing_key.x509

For additional information about signed kernel modules refer to
the documentation privided in the Linux kernel documentation.

https://www.kernel.org/doc/Documentation/module-signing.txt

### DIFF
--- a/config/spl-build.m4
+++ b/config/spl-build.m4
@@ -242,6 +242,8 @@ dnl #
 AC_DEFUN([SPL_AC_RPM], [
 	RPM=rpm
 	RPMBUILD=rpmbuild
+	RPMKEYDIR=$(printf "%s/module" $(pwd))
+	AC_SUBST(RPMKEYDIR)
 
 	AC_MSG_CHECKING([whether $RPM is available])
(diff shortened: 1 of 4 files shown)


behlendorf:
Closing as stale.  This work should be revisited once someone has time to result the remaining issues.

PR #392, opened 2014-09-23
Implement --enable-debuginfo to permit us to force building with debuginfo

ryao:
There are two commits here. The second implements --enable-debuginfo for distributions that build modules without debuginfo. The first is a refactor of --enable-debug based on an observation made during the development of the first that --enable-debug=true would paradoxically become --disable-debug. The order is reversed to make cherry picking easier should anyone want to backport only the change to --enable-debug's argument semantics.

This is the SPL version of zfsonlinux/zfs#2734.


behlendorf:
Closing as unnecessary until I hear otherwise.

ryao:
@behlendorf I am using this at my new job now. It is definitely something useful.

behlendorf:
Closing.  Won't merge.  The SPL is being moved in to the ZFS tree so this will no longer be needed.  The counterpart ZFS patches have been applied to master.

PR #414, opened 2014-11-21
Kmem rework (WIP)

behlendorf:
This pull request is not complete and should __ONLY__ be used on a test system.  I'm opening this pull request to get feedback on the general approach.  It builds on the work @ryao has done in #369.  It has only been lightly testing under CentOS 7 and expect there may still be build failures for other distributions.

@ryao I'd appreciate your feedback on this, particularly on af7f2ab3a8f5ff376f0c343749cc4878f5ed6012 which significantly restructures the code.  This patch stack still needs significant polish but the core ideas are all there.  My commit comments contain notes for some of the areas which I still think need work.

* All the previous patches in #369 have been refresh so they apply cleanly.
* The kmem_alloc/vmem_alloc interfaces have been unified.
* The kmem_alloc/vmem_alloc implementation has been consolidated.
* Split spl-kmem.c in to spl-kmem.c, spl-vmem.c, spl-kmem-cache.c with respective headers.
* Cstyle issues for related source fixes have been fixed.

### DIFF
--- a/include/linux/proc_compat.h
+++ b/include/linux/proc_compat.h
@@ -22,8 +22,8 @@
  *  with the SPL.  If not, see <http://www.gnu.org/licenses/>.
 \*****************************************************************************/
 
-#ifndef _SPL_PROC_H
-#define _SPL_PROC_H
+#ifndef _SPL_PROC_COMPAT_H
+#define _SPL_PROC_COMPAT_H
 
 #include <linux/proc_fs.h>
 
@@ -32,4 +32,4 @@ extern struct proc_dir_entry *proc_spl_kstat;
 int spl_proc_init(void);
 void spl_proc_fini(void);
 
-#endif /* SPL_PROC_H */
+#endif /* SPL_PROC_COMPAT_H */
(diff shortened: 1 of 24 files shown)


behlendorf:
Thanks for the feedback, I've incorporated it and refreshed the patch stack for a fresh round of testing.  As long as nothing unexpected comes out of the testing I'll merge this.

behlendorf:
I've added commit c65cc48 to this patch stack.  It addresses a long standing issue in the spl kmem cache implementation which can result in a deadlock.  @ryao you may be interested in this.

I was able to reliably reproduce this issue while stress testing this patch stack by using a zvol as a swap device.  I choose this workload because it's about the nastiest, most deadlock prone, thing you can do the the VM and the ZFS code.  This kind of test gives me confidences that we haven't missed any of the corner cases.

After applying this fix I've been able to reliably push 100's of GBs to a zvol swap device at an average rate of ~42 MB/s.  Despite the heavy swapping the machine also remains more responsive than with the existing kmem code.

Interestingly enough for this workload the system behaves best when `spl_kmem_cache_slab_limit=0` and the Linux slabs are disabled.  I expect things will behave even better once we move to scatter gather lists for the zio buffers.  But this is an important step in that direction.

behlendorf:
This has been merged as:

9099312 Merge branch 'kmem-rework'

PR #425, opened 2015-01-11
Use current_kernel_time() in the time compatibility wrappers

dweeezil:
Since the Linux kernel's utimens family of functions uses
current_kernel_time(), we need to do the same in the context of ZFS
or else there can be discrepencies in timestamps (they go backward)
if userland code does:

	fd = creat(FNAME, 0600);
	(void) futimens(fd, NULL);

The getnstimeofday() function generally returns a slightly lower time
value.

### DIFF
--- a/include/sys/time.h
+++ b/include/sys/time.h
@@ -54,17 +54,14 @@
 static inline void
 gethrestime(timestruc_t *now)
 {
-	struct timespec ts;
-	getnstimeofday(&ts);
-	now->tv_sec = ts.tv_sec;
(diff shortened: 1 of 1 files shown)


dweeezil:
This fixes zfsonlinux/zfs#3006.  Hopefully it doesn't break anything.

behlendorf:
Nice fix.

behlendorf:
Merged as:

47af4b7 Use current_kernel_time() in the time compatibility wrappers

PR #427, opened 2015-01-23
Fix spl_hostid module parameter

tuxoko:
Currently, spl_hostid module parameter doesn't do anything, because it will
always be overwritten when calling into hostid_read().
Instead, we should only call into hostid_read() when spl_hostid is not zero,
just as the comment describes.


behlendorf:
Looks good to me.

PR #439, opened 2015-03-20
Add mutex_enter_nested() which maps to mutex_lock_nested()

dweeezil:
Also add support for the "name" parameter in mutex_init().  The name
allows for better diagnostics, namely in /proc/lock_stats when
lock debugging is enabled.  Nested mutexes are necessary to support
CONFIG_PROVE_LOCKING. ZoL can use mutex_enter_nested()'s "class" argument
to to convey the locking hierarchy.

### DIFF
--- a/include/sys/mutex.h
+++ b/include/sys/mutex.h
@@ -67,7 +67,7 @@ typedef struct {
 	    (type == MUTEX_ADAPTIVE) ||				\
 	    (type == MUTEX_FSTRANS));				\
 								\
-	__mutex_init(MUTEX(mp), #mp, &__key);			\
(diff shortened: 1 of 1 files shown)


dweeezil:
This is a prerequisite for ZoL to play nice with CONFIG_PROVE_LOCKING.

behlendorf:
I like it.  Let me run it through the buildbots manually to ensure it doesn't break anything on the usual kernels and then I'll get it merged.  It would be nice to have this in place prior to tagging.

behlendorf:
@dweeezil as expected this patch didn't introduce any issues and  has been merged.

PR #440, opened 2015-03-23
Set HAVE_FS_STRUCT_SPINLOCK correctly when CONFIG_FRAME_WARN==1024

dweeezil:
If kernel lock debugging is enabled, the fs_struct structure exceeds the
typical 1024 byte limit of CONFIG_FRAME_WARN and isn't enabled when it
otherwise should be.


dweeezil:
This trivial pull request has been sitting in my to-do list for awhile.  I've increased CONFIG_FRAME_WARN to 2048 on my test systems but it seems that 1024 is rather typical in certain distros (Ubuntu).  There may very well be a few other autos used in the autoconf tests which should be converted to static.

PR #442, opened 2015-04-01
Add hooks to memory allocate function for direct reclaim debugging 

tuxoko:
In order to better catch deadlocks related to direct reclaim, this patch add
hooks to kmem_alloc, kmem_cache_alloc and vmem_alloc in order to simulate
direct reclaim as we desire.

Before using it, you would need to register a callback function. The function
should in turn call to the target shrinker functions.

The whole code is covered with #ifdef KMEM_DEBUG_DIRECT_RECLAIM, so you need
to add it to spl_config.h to enable it.

### DIFF
--- a/include/sys/kmem.h
+++ b/include/sys/kmem.h
@@ -125,8 +125,32 @@ extern unsigned long long kmem_alloc_max;
 extern unsigned int spl_kmem_alloc_warn;
 extern unsigned int spl_kmem_alloc_max;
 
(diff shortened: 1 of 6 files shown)


behlendorf:
At this point we've clearly addressed most/all of the outstanding direct reclaim deadlocks.  However, let's keep this in our back pocket for now since it's can be very useful when developing memory management changes.  I'm just closing the pull request to keep it from being a distraction.

PR #443, opened 2015-04-02
Make tsd_set(key, NULL) remove the tsd entry for current thread

tuxoko:
To prevent leaking tsd entries, we make tsd_set(key, NULL) remove the tsd
entry for the current thread. This is alright since tsd_get() returns NULL
when the entry doesn't exist.

### DIFF
--- a/module/spl/spl-tsd.c
+++ b/module/spl/spl-tsd.c
@@ -391,6 +391,8 @@ tsd_hash_table_fini(tsd_hash_table_t *table)
 	kmem_free(table, sizeof(tsd_hash_table_t));
 }
 
+static void tsd_remove_entry(tsd_hash_entry_t *entry);
+
 /*
  * tsd_set - set thread specific data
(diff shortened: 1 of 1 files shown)


behlendorf:
@tuxoko what about instead of adding `tsd_exit_key()` we update `tsd_set()` such that when passing a NULL value it will destroys the key for this thread.  This should fundamentally solve the issue the same way you've done it, but it has several additional advantages.

* It doesn't require us to extend the thread-specific data API beyond what illumos already supports.  This is good for consistency and let's us avoid Linux specific changes.

* This is consistent with the `tsd_set(key, NULL)` behaviour in illumos in the sense that `tsd_get(key)` will return NULL if either the key doesn't exist or has been set to NULL.  So we can safely destroy the key for the thread any time it's set to NULL.

tuxoko:
@behlendorf 
Yeah, there's nothing preventing us from doing that.

PR #444, opened 2015-04-02
Add crgetzoneid() stub

chrisrd:
Illumos 3897 introduces a dependency on crgetzoneid(). Stub it out until such time as zones are implemented.

References:
  https://github.com/zfsonlinux/zfs/issues/3240

### DIFF
--- a/include/sys/zone.h
+++ b/include/sys/zone.h
@@ -27,7 +27,10 @@
 
 #include <sys/byteorder.h>
 
+#define GLOBAL_ZONEID           0
+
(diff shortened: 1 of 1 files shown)


behlendorf:
Thanks.  I'm happy to get this merged right away so the stub will be available.  Merged as:

c089961 Add crgetzoneid() stub

cburroughs:
>  Stub it out until such time as zones are implemented.

That is a very interesting followup plan.

PR #446, opened 2015-04-02
Don't allow shrinking a PF_FSTRANS context

dweeezil:
Avoid deadlocks when entering the shrinker from a PF_FSTRANS context.

This patch also reverts commit d0d5dd714424365a4da0d887cb641cb2f0ae8844
which added MUTEX_FSTRANS.  Its use has been deprecated within ZFS as it
was an ineffective mechanism to eliminate deadlocks.  Among other things,
it introduced the need for strict ordering of mutex locking and unlocking
in order that the PF_FSTRANS flag wouldn't set incorrectly.

### DIFF
--- a/include/sys/mutex.h
+++ b/include/sys/mutex.h
@@ -32,19 +32,29 @@
 typedef enum {
 	MUTEX_DEFAULT	= 0,
 	MUTEX_SPIN	= 1,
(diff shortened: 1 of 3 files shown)


dweeezil:
This is an adjunct to zfsonlinux/zfs#3225.

PR #450, opened 2015-04-16
vn_getf should return NULL on negative file descriptors

ryao:
C type coercion rules require that negative numbers be converted into
positive numbers via wraparound such that a negative -1 becomes a
positive 1. This causes vn_getf to return a file handle when it should
return NULL whenever a positive file descriptor existed with the same
value. We should check for a negative file descriptor and return NULL
instead.

This was caught by ClusterHQ's unit testing.

Reference:
http://stackoverflow.com/questions/50605/signed-to-unsigned-conversion-in-c-is-it-always-safe

Signed-off-by: Andriy Gapon <avg@FreeBSD.org>

### DIFF
--- a/module/spl/spl-vnode.c
+++ b/module/spl/spl-vnode.c
@@ -648,6 +648,16 @@ vn_getf(int fd)
 	vnode_t *vp;
 	int rc = 0;
 
+	/*
+	 * Negative file descriptors passed to fget() will be implicitly
+	 * converted into positive file descriptors by C's type coercion rules
+	 * such that the value becomes fd + UINT_MAX + 1. That is wrong, so we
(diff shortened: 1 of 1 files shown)


sempervictus:
Just read through a bunch of the reference material, all of it a strong supporting argument for the postulate that "the devil is in the details." Any chance we could push up some pointer to our contextual understanding of the reason for why we got a negative FD in the first place? My (meager) understanding of the subject matter is that POSIX systems issue negative FDs in error conditions...

ryao:
@sempervictus A negative fd was intentionally passed via cleanup_fd in the /dev/zfs API to test behavior for sanity. FreeBSD's ZFS driver passed while ZoL failed.

PR #453, opened 2015-05-14
Make taskq_wait() block until the queue is empty

behlendorf:
Under Illumos taskq_wait() returns when there are no more tasks
in the queue.  This behavior differs from ZoL and FreeBSD where
taskq_wait() returns when all the tasks in the queue at the
beginning of the taskq_wait() call are complete.  New tasks
added whilst taskq_wait() is running will be ignored.

This difference in semantics makes it possible that new subtle
issues could be introduced when porting changes from Illumos.
To avoid that possibility the taskq_wait() function is being
updated such that it blocks until the queue in empty.  The
previous behavior is still available via the taskq_wait_all()
interface.

Signed-off-by: Chris Dunlop <chris@onthe.net.au>

### DIFF
--- a/include/sys/taskq.h
+++ b/include/sys/taskq.h
@@ -119,7 +119,7 @@ extern void taskq_init_ent(taskq_ent_t *);
 extern taskq_t *taskq_create(const char *, int, pri_t, int, int, uint_t);
 extern void taskq_destroy(taskq_t *);
 extern void taskq_wait_id(taskq_t *, taskqid_t);
-extern void taskq_wait_all(taskq_t *, taskqid_t);
+extern void taskq_wait_outstanding(taskq_t *, taskqid_t);
 extern void taskq_wait(taskq_t *);
 extern int taskq_cancel_id(taskq_t *, taskqid_t);
 extern int taskq_member(taskq_t *, void *);
--- a/module/spl/spl-taskq.c
+++ b/module/spl/spl-taskq.c
@@ -327,6 +327,33 @@ taskq_find(taskq_t *tq, taskqid_t id, int *active)
 	return (NULL);
 }
 
+/*
+ * Theory for the taskq_wait_id(), taskq_wait_outstanding(), and
+ * taskq_wait() functions below.
+ *
+ * Taskq waiting is accomplished by tracking the lowest outstanding task
+ * id and the next available task id.  As tasks are dispatched they are
+ * added to the tail of the pending, priority, or delay lists.  As worker
+ * threads become available the tasks are removed from the heads of these
+ * lists and linked to the worker threads.  This ensures the lists are
+ * kept sorted by lowest to highest task id.
+ *
+ * Therefore the lowest outstanding task id can be quickly determined by
+ * checking the head item from all of these lists.  This value is stored
+ * with the taskq as the lowest id.  It only needs to be recalculated when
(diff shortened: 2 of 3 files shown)


behlendorf:
@chrisrd what do you think about this?  We update the `taskq_wait()` implementation to have the illumos semantics.  My only slightly concern with this is it _might_ end up have a performance impact.  However, that could be pretty easily mitigated because we could also update critical areas to use `taskq_wait_all()` to get the previous semantics.

chrisrd:
@behlendorf the code looks good.

The `taskq_wait_all()` name seems at odds with it's functionality: to my mind `taskq_wait()` (now) waits for "all" tasks to complete, whereas the renamed `task_wait_all()` is waiting for all outstanding tasks when the function was called, so possibly `taskq_wait_outstanding()`?

Performance wise, what's the worst that could happen? I guess we could potentially not return for a very long time if tasks are constantly getting queued, e.g. due to memory pressure. If we have any queuing of tasks in the freeing memory path that could potentially end up as a real or apparent deadlock. That would be pretty bad for performance!

behlendorf:
@chrisrd agreed, I'm not to happy with `taskq_wait_all()` either, it was a poor choice of name.  The good news is that nothing in ZFS uses it since it was primarily added for testing purposes  in the SPLAT.  So, if we're going to change it now is the time to do it.  I like `taskq_wait_outstanding()` let me refresh this accordingly.

As for performance I don't think we'll encounter anything too serious because that most likely would already have been exposed under illumos.  But it's just something to keep in mind and I wanted to mention!  It's probably more likely that we'll just get some inexplicable and unlikely bugs fixed.

behlendorf:
@chrisrd refreshed.  Locally I've run the updated SPL through the buildbot and everything passed.  I'll test the refreshed patch again but assuming nothing comes up I think we can merge this.  Can you look over the refreshed version, it just includes the function name change.

chrisrd:
Looks good to me!

behlendorf:
Interestingly, this patch appears to cause the occasional `cannot export 'tank': pool is busy` error in the automated `zimport.sh` tests.  I'm fairly sure retrying the command would succeed so I wouldn't categorize this as a regression.  I suspect we're just blocking slightly longer now with a reference on the spa, which arguably is safer.  I'll try adjusting the test script to account for this and verify it's a transient problem before merging this.

behlendorf:
Adding a delay between the immediate import/export appears to avoid the spurious busy error.

ryao:
@behlendorf  If we import the Illumos taskq code, we can avoid regressions like this by getting the exact Illumos behavior. I have a proof of concept in a local branch that I can push if you are interested. It depended on the kmem-rework, which has since been merged.

behlendorf:
@ryao I'm not sure I'd categorize this as a regression.  The documentation was a bit ambiguous about this and FreeBSD implemented the same way we did.  Frankly, I'd prefer to keep the existing behavior which is probably better for performance but I don't like the risk of accidental bugs being introduced due to a behavior differences.

I'm open to the idea of moving to the Illumos implementation but I'd need to be convinced of a few things.  Off the top of my head:
* It performs as well, or better, than the existing implementation.
* taskq_dispatch() can be safely called in an atomic context.

behlendorf:
Closing, replaced by #455.

PR #460, opened 2015-07-01
Add a script to display SPL slab cache statistics

bprotopopov:
I thought this might be useful when looking for the info on ZFS/SPL related memory consumption in Linux. The sources I have consulted included:
- /proc/meminfo (there is a number of commands that interpret and display this info already)
- /proc/slabinfo (there is a nice slabtop command to display this)
- /proc/spl/kmem/slab (I did not know of a way to show this into in a convenient way)

I was thinking of using python curses to present the output in a way slabtop does it, but did not have time as of yet. Also, I am currently showing a subset of the info from the /proc/spl/kmem/slab. Room for improvement :)


behlendorf:
Definitely useful and something we can build on, thanks for opening the pull request.  I didn't run in to any problems when testing the script so I've merged it.  Thanks for remembering to add it to the packaging!

5578f58 Add a script to display SPL slab cache statistics

PR #461, opened 2015-07-09
Enforce build requirements for deb-utils

pfelecan:
IMHO, when a prerequisite for deb-utils is missing, such as 'alien'
it's better to fail during the configuration phase than when building.

### DIFF
--- a/config/spl-build.m4
+++ b/config/spl-build.m4
@@ -270,6 +270,7 @@ AC_DEFUN([SPL_AC_DPKG], [
 	],[
 		HAVE_DPKG=no
 		AC_MSG_RESULT([$HAVE_DPKG])
+		AC_MSG_ERROR(["please install the $DPKG package and re-run configure"], [1])
 	])
 
 	AC_MSG_CHECKING([whether $DPKGBUILD is available])
@@ -281,6 +282,7 @@ AC_DEFUN([SPL_AC_DPKG], [
 	],[
 		HAVE_DPKGBUILD=no
 		AC_MSG_RESULT([$HAVE_DPKGBUILD])
+		AC_MSG_ERROR(["please install the $DPKGBUILD package and re-run configure"], [1])
 	])
 
 	AC_SUBST(HAVE_DPKG)
@@ -309,6 +311,7 @@ AC_DEFUN([SPL_AC_ALIEN], [
 	],[
 		HAVE_ALIEN=no
 		AC_MSG_RESULT([$HAVE_ALIEN])
+		AC_MSG_ERROR(["please install the $ALIEN package and re-run configure"], [1])
 	])
 
 	AC_SUBST(HAVE_ALIEN)


pfelecan:
@behlendorf I totally agree with you but this wouldn't be necessary if the Debian packaging stuff, as proposed in the pull request #289, were present; in that case, the build prerequisites are intrinsically looked for by the Debian build framework, i.e. control and dh*. There is an rpm directory, why not a debian one. the zfs-pkg is, from my point of view a palliative By smoothly integrating Debian helps the acceptance of the packages in the main stream and, thus, its usage in the corresponding ecosystem and, eventually, face competition with other, less mature, file-systems, e.g. btrfs.

FransUrbo:
> There is an rpm directory, why not a debian one.
> 
My point exactly! The "excuse" have been that "Debian don't like it when you provide a debian dir upstream". But there really IS no such rule (as far as I know - haven't studied the "Debian GNU/Linux Packaging Guidelines" in years).

But since I'm one of the three Debian GNU/Linux Developers in charge of the official (and the one in charge of the unofficial - The ZoL packages) packages for Debian GNU/Linux, _I_ can say there's no problem here. And I've talked to the other two guys, and they don't have any major issues with it either.
> the zfs-pkg is, from my point of view a palliative
> 
Agree! [It] just complicates issues for me and for everyone else.

pfelecan:
Can this be reconsidered and pull the request #289 ? Probably amended with the evolution of Debian releases.
What thinks @behlendorf ?
I'm eager to help if needed.

behlendorf:
> The "excuse" have been that "Debian don't like it when you provide a debian dir upstream". 

It's my understanding that this is something which the Debian project discourages, but doesn't prohibit, upstream repositories from doing.  This is supposed to make it easier for the downstream maintainers.  In practice I don't know enough about Debian to say if that's true or not.

https://wiki.debian.org/UpstreamGuide#Pristine_Upstream_Source

My understanding is that including a debian directory is a problem because, unlike an rpm.spec file, there's no way to write a single debian directory which works for multiple distributions or even releases of the same distribution.  So even if we wanted to we couldn't provide one that _just works_.

@FransUrbo would it be possible for us to provide a debian directory which for exampled worked for both Wheezy and Sqeeeze.  I image it's technically achievable using `.in` files but its that desirable?

FransUrbo:
> @FransUrbo would it be possible for us to provide a debian directory which for exampled worked for both Wheezy and Sqeeeze.
> 
Yes. There's _very little_ differences between the two, so it would be fairly easy to do. We'd also need a new, empty changes file (so it get's built every time "a user" builds the packages - we need the GIT sha code etc).

All fairly easy to do, with .in files as you say. I already do some of that (in the pkg-{zfs,spl} repositories), to account for stuff like the SO versioning and a couple of other things.

Extending that to take into account Wheezy, Jessie and the different Ubuntu versions shouldn't be to difficult. Since a lot of people run my packages on different Ubuntu versions instead of the one @dajhorn is providing, I bet they would be very grateful to for something like that.


Since Debian GNU/Linux package maintainers "is encouraged" to have the debian directory in the Debian GNU/Linux git for the package (getting the source from one location and the debian dir from another), it wouldn't take to much work to "augment" the upstream (our) debian directory with the changes etc that would be needed/wanted for the official packages. Such as a proper changes file - the only one I can think of of-hand. That MUST be in a continuing state for the packages uploaded to the Debian GNU/Linux FTP repo…


As I said, much of this already exists in the packages I build for ZoL so improve upon it a little more to have it in the **zfs** repo wouldn't take to much work.

pfelecan:
Good. Can we advance on this?
I can test and adapt on Jessie and Snatch, i.e., stable and testing.

behlendorf:
Closing, I'd prefer not to merge this.  Failing at configure time when Debian packages aren't going to be build will result in quite a few surprised users.  I think we'll just have to be OK with logging the failure at build time.

PR #465, opened 2015-07-23
Remove skc_ref from alloc/free paths     

behlendorf:
As described in spl_kmem_cache_destroy() the ->skc_ref count was
added to address the case of a cache reap or grow racing with a
destroy.  They are not strictly needed in the alloc/free paths
because consumers of the cache are responsible for not using it
while it's being destroyed.
    
Removing this code is desirable because there is some evidence that
contention on this atomic negative impacts performance on large-scale
NUMA systems.


dweeezil:
@behlendorf LGTM and largely matches the change I made myself prior to your posting of this one.

behlendorf:
Merged as:

4699d76 Remove skc_ref from alloc/free paths

PR #467, opened 2015-07-27
Remove needfree, desfree, lotsfree #defines

behlendorf:
This patch reverts 77ab5dd.  This is now possible because upstream has
refactored the ARC is such a way that these values are only used in a
few key places.  Those places have subsequently been updated to use
the default illumos setting but also exposed as module option to
allow tuning if required.

Issue zfsonlinux/zfs#3637


behlendorf:
Merged as:

8ac6ffe Remove needfree, desfree, lotsfree #defines

PR #468, opened 2015-07-30
Restructure uio to accommodate bio_vec

tuxoko:
Starting from Linux 4.1, bio_vec will be allowed to pass into filesystem via
iter_read/iter_write, so we add a bio_vec field in uio_t to hold it, and use
UIO_BVEC in segflg to determine which "vec".

Also, to be consistent to newer kernel, we make iovec and bio_vec immutable,
and make uio act as an iterator with the new uio_skip field.

### DIFF
--- a/include/sys/uio.h
+++ b/include/sys/uio.h
@@ -1,6 +1,7 @@
 /*****************************************************************************\
  *  Copyright (C) 2007-2010 Lawrence Livermore National Security, LLC.
  *  Copyright (C) 2007 The Regents of the University of California.
+ *  Copyright (c) 2015 by Chunwei Chen. All rights reserved.
  *  Produced at Lawrence Livermore National Laboratory (cf, DISCLAIMER).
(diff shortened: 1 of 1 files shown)


l1k:
Tested-by: Lukas Wunner <lukas@wunner.de>

ryao:
I have tested this too by applying it to spl-0.6.4.2. It works fine.

infowolfe:
working with sys-kernel/aufs-sources-4.1.5

tuxoko:
Added ```#include <linux/blkdev.h>```

kernelOfTruth:
@tuxoko I'll test this updated patch and https://github.com/zfsonlinux/zfs/pull/3651 once I get the opportunity

Thanks !

tuxoko:
@kernelOfTruth 
This update only fix compile error for older kernel. So it shouldn't matter.

PR #471, opened 2015-08-18
Include other sources of freeable memory in the freemem calculation

behlendorf:
This prevents ARC collapse when non-ZFS filesystems, the block layer or
other consumers use a lot of reclaimable memory.


dweeezil:
I updated the commit comment a bit in 35c190cdd52c261233b127ade23b50958fea685e.

behlendorf:
Thanks.  Overnight testing went well and didn't uncover any unexpected issues.  Merged to master with the updated comment.

851a549 Include other sources of freeable memory in the freemem calculation

PR #472, opened 2015-08-27
Create a new thread during recursive taskq dispatch if necessary

dweeezil:
When dynamic taskq is enabled and all threads for a taskq are occupied,
a recursive dispatch can cause a deadlock if calling thread depends on
the recursively-dispatched thread for its return condition.

This patch attempts to create a new thread for recursive dispatch when
none are available.

### DIFF
--- a/module/spl/spl-taskq.c
+++ b/module/spl/spl-taskq.c
@@ -448,8 +448,8 @@ taskq_wait(taskq_t *tq)
 }
 EXPORT_SYMBOL(taskq_wait);
 
-int
-taskq_member(taskq_t *tq, void *t)
+static int
+taskq_member_impl(taskq_t *tq, void *t)
 {
 	struct list_head *l;
 	taskq_thread_t *tqt;
@@ -457,8 +457,8 @@ taskq_member(taskq_t *tq, void *t)
 
 	ASSERT(tq);
 	ASSERT(t);
+	ASSERT(spin_is_locked(&tq->tq_lock));
 
-	spin_lock_irqsave(&tq->tq_lock, tq->tq_lock_flags);
 	list_for_each(l, &tq->tq_thread_list) {
 		tqt = list_entry(l, taskq_thread_t, tqt_thread_list);
 		if (tqt->tqt_thread == (struct task_struct *)t) {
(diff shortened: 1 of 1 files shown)


dweeezil:
I ran into this issue while working on the TRIM code from Nexenta.
Its spa code which dispatches the underlying per-vdev trims to the
system taskq.  Unfortunately, the existing vdev file code also uses the
system taskq so sometimes, those dispatches are executed in the existing
one-and-only system taskq thread.

This all works when dynamic taskqs are not enabled because there's almost
always a thread available to handle the recursive dispatch.  When dynamic
taskqs are enabled, however, there are frequently no threads available
and if all the existing ones are used, the recursive dispatch (the vdev
file operation in this case) never completes.

As a bit of further background, the actual symptom observed within ZFS
is a stuck zio which happens to be a twice-removed child from a NULL
zio and it was rather difficult to trace back to its origin.

I'm also wondering if this is not really a bug and that the users of
taskqs are supposed to know this is a Bad Thing.  The zio code has a
few places where it performs a similar check but, it seems for a wholly
different purpose.

behlendorf:
This is absolutely something the users of taskqs need to be aware of!  But I suspect your right this never turned up on Illumos because their system taskq see a lot more activity than ours.  Plus they do behave a little differently so I could easily see an unrelated system task coming along and knocking things free so to speak.

The proposed fix looks good to me but I'd extend it to include the `taskq_dispatch_ent()` and `taskq_dispatch_delay()` functions too.  There's no reason a similar situation couldn't occur there.

This patch doesn't cover the case where we've already created the maximum number of threads and they're all blocking.  But frankly there's nothing we can do about that ensure that taskq is large enough that this can never happen.

dweeezil:
OK, I kind of figured it up to the users of taskqs to be aware of this.  On the other hand, that does seem to be worth working around.

I was planning on extending it to the other two functions if I got decent feedback, so I'll update the pull request.

More than anything else, this behavior made me wonder if there are any other dark corners of ZFS which might fall victim to it.

behlendorf:
Thanks for refreshing this.  It LGTM, the only minor change I made was moving for forward declaration of to the top of the file with the others.

As for other dark corners I suspect this might resolve an issue we hit in the zfs-test suite where pools are created within pools.  But I haven't verified that.

behlendorf:
Reopening and reverting the patch.  Subsequent additional testing uncovered an issue.

```
 VERIFY3(0 == tq->tq_nspawn) failed (0 == -1)
```

taskq_thread_spawn() needs to be called under the spin lock so tq_nspawn is updated safely.  That's a little problematic since taskq_member() can't be called under the lock since it want to take it.  We'll need to slightly rejigger the code to handle this.  My first thought in to add a static `taskq_member_impl()` function for internal use which requires the caller to already have the lock.  That way this can all be done under the lock.  That should also remove the need for a local variable entirely.

dweeezil:
@behlendorf Re-working as you suggested.  I was so keen on re-using the existing ```taskq_member()``` that it never dawned on me to convert it into a wrapper which didn't require the lock.

behlendorf:
I like this better and no problem were observed in testing.  Merged!

PR #474, opened 2015-09-07
Disable direct reclaim in taskq worker threads on Linux 3.9+

ryao:
Illumos does not have direct reclaim and code run inside taskq worker
threads is not designed to deal with it. Allowing direct reclaim inside
a worker thread can therefore deadlock. We set PF_MEMALLOC_NOIO through
memalloc_noio_save() to indicate to the kernel's reclaim code that we
are inside a context where memory allocations cannot be allowed to block
on filesystem activity.

### DIFF
--- a/module/spl/spl-taskq.c
+++ b/module/spl/spl-taskq.c
@@ -817,6 +817,10 @@ taskq_thread(void *args)
 	tq = tqt->tqt_tq;
 	current->flags |= PF_NOFREEZE;
 
+	#if defined(PF_MEMALLOC_NOIO)
+	(void) memalloc_noio_save();
+	#endif
+
 	sigfillset(&blocked);
 	sigprocmask(SIG_BLOCK, &blocked, NULL);
 	flush_signals(current);


ryao:
@behlendorf I realize that you want to try a more surgical approach, but I am opening this so there is something to track this and also to provide myself with an easy way to supply a patch to people who need swap working right now.

That said, I have made a revision since the earlier iteration that set PF_FSTRANS after realizing that PF_FSTRANS will change the kmem allocator to make KM_SLEEP into KM_PUSHPAGE and set PF_MEMALLOC_NOIO on Linux 3.9+. All that we should need (in theory) is PF_MEMALLOC_NOIO. Anything that is KM_SLEEP in a taskq on Illumos should be safe in a taskq on Linux.

kernelOfTruth:
@ryao Great !

I had swap support on my mind the last few days,

glad it'll work with this :+1:

behlendorf:
@ryao I don't think this is an unreasonable approach since it specifically targets IO during direct reclaim and leaves all the other reclaim strategies intact.  It would be great to get confirm this does resolve the outstanding swap issues.

ryao:
@behlendorf I will try to do some stress tests on my laptop today and let you know.

ryao:
Preliminary test results of swap on a zvol on my laptop with this patch applied to HEAD strongly suggests that this prevents deadlocks.

The methodoogy is as follows:

0. Setup a zvol with primarycache=metadata secondarycache=none sync=always and logbias=throughput as swap so that paged out memory is evicted quickly.
1. Start 8 instances of `python -c 'print 2**10**100' &` to gobble up system memory.
2. Wait for system memory to be exhausted such that page out occurs and pagefaults effectively lock up Xorg.
3. Attempt to regain control of the system via Magic Sysreq key by killing processes via either manual OOM execution or killing all processes.
4. Verify that the pool still works.

If we deadlock, step 3 is not possible because the ZIO pipeline has stalled and the only solution is to force a reboot. The reason Xorg and other things lock up is because the python processes are starving interactive processes for unpaged memory by consuming memory faster than pageout and accessing it all in a loop.

behlendorf:
No unexpected issues observed during testing.  Because this is a minimal change designed to improve the robustness of the system it has been merged as:

d4bf6d8 Disable direct reclaim in taskq worker threads on Linux 3.9+

PR #487, opened 2015-10-24
Don't call kmem_cache_shrink from shrinker

tuxoko:
Linux slab will automatically free empty slab when number of partial slab is
over min_partial, so we don't need to explicitly shrink it. In fact, calling
kmem_cache_shrink from shrinker will cause heavy contention on
kmem_cache_node->list_lock, to the point that it might cause __slab_free to
livelock (see zfsonlinux/zfs#3936)


behlendorf:
LGTM, nice fix.  Merged.

PR #488, opened 2015-10-24
sys/sysmacros.h: Fix P2ROUNDUP_TYPED to not trigger integer overflow

perfinion:
The original P2ROUNDUP_TYPED macro contains a -x which triggers PaX's
integer overflow detection for unsigned integers. Replace the macro with
an equivalent version that does not trigger the overflow.

Signed-off-by: Jason Zaman <jason@perfinion.com>

Fixes issue https://github.com/zfsonlinux/zfs/issues/2505


perfinion:
I updated the commit message after discussion with @ryao and included the proof from the original github issue

PR #491, opened 2015-11-01
Provide kstat for taskqs in /proc/spl/taskqs

dweeezil:
I'm posting this as a WIP for further commentary and feedback.  I think the output is potentially useful as it stands but there's plenty that could be changed, added or deleted..  The main goal at the moment is to provide a way to help track down issues regarding dynamic taskqs but the information provided could certainly be useful in other contexts.

Among other things, I've thought about adding a nicer header line and grouping the columns, for example, into a "tqent" section, and others.  There are also likely some non-useful columns at the moment and maybe some others which ought to be added.

On a smaller test system running a bunch of ```zfs send``` and other miscellaneous things running current master code with dynamic taskqs enabled, here's a sample of the output:
```
taskq                       act  nthr  spwn  maxt   pri  mina         maxa  cura      flags
spl_kmem_cache/0              0     1     0     4   100    32   2147483647    32   80000005
spl_system_taskq/0            1     1     0    64   100     8   2147483647     8   80000005
	traverse_prefetch_thread [zfs](0xffff8800c690e420)
spl_dynamic_taskq/0           0     1     0     1   100     8   2147483647     8   80000001
dbu_evict/0                   0     1     0     1   120     0            0     0   80000000
arc_prune/0                   0     1     0     8   120     8   2147483647     8   80000005
z_unmount/0                   0     1     0     1   120     1            8     1   80000001
z_null_iss/0                  0     1     0     1   100    50   2147483647     0   80000004
z_null_int/0                  0     1     0     1   100    50   2147483647     0   80000004
z_rd_iss/0                    0     1     0     8   100    50   2147483647     0   80000004
z_rd_int_0/0                  0     1     0    12   100    50   2147483647     0   80000004
z_rd_int_1/0                  0     1     0    12   100    50   2147483647     0   80000004
...
z_rd_int_1/1                  0     1     0    12   100    50   2147483647     0   80000004
z_rd_int_2/1                  0     1     0    12   100    50   2147483647     0   80000004
z_rd_int_3/1                  0     1     0    12   100    50   2147483647     0   80000004
z_rd_int_4/1                  1     1     0    12   100    50   2147483647     0   80000004
	zio_execute [zfs](0xffff8802256fc1a0)
z_rd_int_5/1                  0     1     0    12   100    50   2147483647     0   80000004
z_rd_int_6/1                  0     1     0    12   100    50   2147483647     0   80000004
z_rd_int_7/1                  0     1     0    12   100    50   2147483647     0   80000004
z_wr_iss/1                    0     1     0     6   101    50   2147483647     0   8000000c
z_wr_iss_h/1                  0     1     0     5   100    50   2147483647     0   80000004
...
metaslab_group_taskq/1        0     1     0     4   100    10   2147483647     0   8000000c
metaslab_group_taskq/2        0     1     0     4   100    10   2147483647     0   8000000c
z_iput/1                      0     1     0     8   120    64   2147483647    64   80000005
```
The "/X" in the name of the taskq is what I'm calling the "instance number".  It's a monotonically increasing integer to help distinguish otherwise identically-named taskqs.

### DIFF
--- a/module/spl/spl-proc.c
+++ b/module/spl/spl-proc.c
@@ -29,6 +29,7 @@
 #include <sys/kmem.h>
 #include <sys/kmem_cache.h>
 #include <sys/vmem.h>
+#include <sys/taskq.h>
 #include <linux/ctype.h>
 #include <linux/kmod.h>
 #include <linux/seq_file.h>
@@ -49,6 +50,8 @@ static struct ctl_table_header *spl_header = NULL;
 static struct proc_dir_entry *proc_spl = NULL;
 static struct proc_dir_entry *proc_spl_kmem = NULL;
 static struct proc_dir_entry *proc_spl_kmem_slab = NULL;
+static struct proc_dir_entry *proc_spl_taskq_all = NULL;
+static struct proc_dir_entry *proc_spl_taskq = NULL;
 struct proc_dir_entry *proc_spl_kstat = NULL;
 
 static int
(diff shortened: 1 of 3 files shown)


tuxoko:
@dweeezil 
I like this and I think we would definitely need this. But I think we probly also need number of tasks in the wait list.

dweeezil:
@tuxoko Thanks for the feedback.  I'll look it over and work on it over the weekend.

dweeezil:
I'm going to try to get this refreshed this weekend based on the feedback I've gotten so far.

dweeezil:
The latest refresh (51b428b) adds another kstat.  Now there are both ```/proc/spl/taskq-all``` and ```/proc/spl/taskq```.  The former will display all taskqs whereas the latter will only display those with active threads or with pending work.

Here's an example of a somewhat overloaded write issue taskq with several threads running and one in the pending list.  The pid is listed in square brackets for active threads.

```
root@zfs-dev:~# cat /proc/spl/taskq
taskq                       act  nthr  spwn  maxt   pri  mina         maxa  cura      flags
z_wr_iss/0                    7    59     1    60   101    50   2147483647     0   8000000c
	active: [20115]zio_execute [zfs](0xffff885e2e6f1a40) [65185]zio_execute [zfs](0xffff885e2e6f7620)
	        [20282]zio_execute [zfs](0xffff885e2e6f71c0) [20328]zio_execute [zfs](0xffff885e885ab020)
	        [20229]zio_execute [zfs](0xffff885e2e6f7a80) [19767]zio_execute [zfs](0xffff885e885ac1a0)
	        [20347]zio_execute [zfs](0xffff885e885ac600)
	pend: zio_execute [zfs](0xffff885e2e6f7620)
```
I think the only important thing missing now is information regarding the wait queues.  I'll be adding something shortly for those.

dweeezil:
Added output of the ```tq_wait_waitq``` wait queue.  The work waitq isn't terribly interesting, at least under normal circumstances.  Here's a sample grabbed while running splat showing the pids of the waiters:
```
taskq                       act  nthr  spwn  maxt   pri  mina         maxa  cura      flags
front/0                       3     3     0     3   120    50   2147483647    50   80000001
	active: [55104]splat_taskq_test6_func [splat](0xffff881f63e83d38) [55105]splat_taskq_test6_func [splat](0xffff881f63e83d48)
	        [55103]splat_taskq_test6_func [splat](0xffff881f63e83d78)
	pend: splat_taskq_test6_func [splat](0xffff881f63e83d58) splat_taskq_test6_func [splat](0xffff881f63e83d68)
	prio: splat_taskq_test6_func [splat](0xffff881f63e83d88) splat_taskq_test6_func [splat](0xffff881f63e83d98)
	wait: 55073
```

dweeezil:
Although I think this is getting to be pretty useful as-is, I'm thinking that with this bit of infrastructure, it would be pretty cool to add a new form of ```taskq_dispatch()``` and friends which would register a callback to be used by the taskq kstat to decode the argument of the spawned function.  In particular, it could be very useful to do at least a partial decode of the zio in ```zio_execute()```.

tuxoko:
The problem with ```tq_wait_waitq``` is that in many cases the waiter for a task is waiting on a task specific cv rather than the waitq.

@dweeezil 
Are you aware of any taskq deadlock after my remove ```taskq_member_impl``` patch? I've been having this idea of a kicker thread, that would periodically scan the pending list head. And if it sees a task waited too long, it would "kick" the taskq by spawning a new thread.

tuxoko:
@dweeezil 
While decoding the zio sounds like a cool idea. But I don't think you can safely do that because zio will be freed asynchronously from the one doing the stat.

behlendorf:
> Although I think this is getting to be pretty useful as-is...

Absolutely, I really like the way this is turning out.  I can see it helping tremendously with debugging and performance analysis.  It would be great to finalize your existing implementation so it can be applied to the spl and used as a basis for future improvements.

> add a new form of taskq_dispatch() and friends which would register a callback 

That's an interesting idea.  I think it _could_ be done safely but it would need to be the responsibility of the callback to ensure the memory wasn't freed either using a reference count on the object or some kind of locking.  For zio's in particular it doesn't look like there's an existing way to prevent it from racing with `zio_destroy()`.

> Are you aware of any taskq deadlock after my remove taskq_member_impl patch.

@tuxoko I haven't heard of any more instances after applying your patch.  For the moment it doesn't look like we're going to need a kicker thread, but let's keep that idea in our back pocket just in case.

module/spl/spl-proc.c
```diff
@@ -216,6 +219,151 @@ proc_dohostid(struct ctl_table *table, int write,
...
+	wait_queue_t *wq;
+	struct task_struct *tsk;
+	taskq_ent_t *tqe;
+	char name[100];
+	struct list_head *lheads[4], *lh;
+	static char *list_names[sizeof(lheads) / sizeof(lheads[0])] =
+	    {"pend", "prio", "dely", "wait" };
+	int i, j, have_lheads = 0;
+
+	spin_lock(&tq->tq_lock);
```
tuxoko: I think you should use spin_lock_irqsave/spin_unlock_irqrestore

module/spl/spl-proc.c
```diff
@@ -216,6 +219,151 @@ proc_dohostid(struct ctl_table *table, int write,
...
+
+	lheads[0] = &tq->tq_pend_list;
+	lheads[1] = &tq->tq_prio_list;
+	lheads[2] = &tq->tq_delay_list;
+	lheads[3] = &tq->tq_wait_waitq.task_list;
```
tuxoko: waitq has its own spinlock to protect the list.

dweeezil:
@tuxoko I wasn't sure about the inner lock so I used irqsave/restore for safety to begin with.

While troubleshooting a reproducible hang, I had already changed my use of ```tq->tq_lock_flags``` to a local before seeing #506.  I don't have my small test system booted right now but I've isolated the problem to a high frequency of taking and releasing ```tq->tq_lock``` with irqsave/restore (and nothing to do with the lock taken on the waitq).  I'm going to try #506 ASAP to see if it solves the problem.

The test is performed on a small test system with a pool consisting of a partition on an SSD and simply doing (dynamic taskqs enabled):
```
while :; do
  zpool import tank
  zpool export tank
done
```
and
```
while :; do
  cat /proc/spl/taskq > /dev/null
done
```
I designed this test because in order to stress the waitq locking and also cause the kstat to show a non-empty waitq more often.

dweeezil:
I'll also note that I'm letting a few commits stack up in this PR for the time being but will get them squashed with a good commit comment once I'm happy with it.

module/spl/spl-taskq.c
```diff
@@ -1081,6 +1108,10 @@ taskq_destroy(taskq_t *tq)
 		task_free(tq, t);
 	}
 
+	down_write(&tq_list_sem);
+	list_del(&tq->tq_taskqs);
+	up_write(&tq_list_sem);
```
tuxoko: This cannot be in spinlock context.

module/spl/spl-proc.c
```diff
@@ -476,6 +695,20 @@ spl_proc_init(void)
 		goto out;
 	}
 
+	proc_spl_taskq_all = proc_create_data("taskq-all", 0444,
+		proc_spl, &proc_taskq_all_operations, NULL);
+	if (proc_spl_taskq_all == NULL) {
+		rc = -EUNATCH;
```
tuxoko: The use of EUNATCH is a really strange choice. I know you just copy the above ones. But perhaps we should change them all to ENOMEM.

module/spl/spl-taskq.c
```diff
@@ -54,6 +54,10 @@ EXPORT_SYMBOL(system_taskq);
 static taskq_t *dynamic_taskq;
 static taskq_thread_t *taskq_thread_create(taskq_t *);
 
+/* List of all taskqs */
+struct list_head tq_list;
+struct rw_semaphore tq_list_sem;
```
tuxoko: You can just ```LIST_HEAD(tq_list); DECLARE_RWSEM(tq_list_sem);``` so you don't need to bother init it some where else.

module/spl/spl-taskq.c
dweeezil: Oops, sleep during spinlock is not good.

module/spl/spl-proc.c
dweeezil: Seems reasonable to me but possibly in its own commit.

behlendorf: Yes, in a follow up commit.

module/spl/spl-taskq.c
behlendorf: I'd also recommend removing it from the list after `taskq_wait()` above and before the spin lock.  We really don't want the taskq_t to be findable while it's being destroyed.

dweeezil:
This looks pretty good after the changes in be27b78.  After a few more hours of stress testing, I'll get the commits squashed and rebased and add a proper commit comment.

behlendorf:
@dweeezil sounds good.  Let me know when you're happy with them and I'll look it over again so we can get it merged.

dweeezil:
I've thrashed this for about 26 hours on a rather large system and it held up perfectly well.  I'll get the commits squashed and add an appropriated comment and then it ought to be ready to merge.

behlendorf:
@dweeezil it sounds like this is very close to ready.

dweeezil:
@behlendorf I'll get this finalized within the next day.

behlendorf:
@dweeezil that would be great.  I've just merged several of the outstanding taskq patches so you'll definitely want to rebase on master.

dweeezil:
@behlendorf Rebased to master, which involved updating a few things, and added a more complete commit comment.  Compiles but untested at the moment; I pushed it so I can easily grab it on my test system.

dweeezil:
It looks like the rebase went OK.  I stressed it for awhile this morning and it appears to be good.  This should be ready to merge now.

behlendorf:
Aside from the dely question this looks good to me, I didn't have any issues testing it.  And for the record is it very cool it's awesome to be able to see at a glance what the taskqs are doing.

dweeezil:
@behlendorf Yes, the "dely" was for alignment (feel free to change it to "delay").  I've been using this patch on all my recent test rigs and it's very very handy to get this visualization of the taskqs.

behlendorf:
@dweeezil I'm OK either, I just wasn't sure because you didn't also shorten "active".  I don't feel strongly either way, I'd go "delay" if I had to choose, but I think we should consistent.    What do you prefer?

dweeezil:
@behlendorf I just pushed it with "delay".  For reasons that escape me now, I was mentally grouping the other lists (not "active") together and thinking it would be good to keep their alignment consistent.  In practice, I don't think it matters.  I hated "dely" from the get-go.

behlendorf:
OK then, "delay" it is.  LGTM, I've merged your patch to master.

PR #493, opened 2015-11-05
Vnode improvements

ryao:
The vnode code has redundant mutexes and atomics.


behlendorf:
Since this is just cleanup which shouldn't have any real impact on performance and this is all code I'd like to return (see #497) I'd prefer not to merge this.

PR #494, opened 2015-11-05
Support MUTEX_SPIN

ryao:
I implemented this so that I could use spinlocks with condition variables when prototyping a solution to the areleasef()/getf() race. However, the simpler and more elegant solution of #492 rendered the use of condition variables unnecessary. This is unnecesary as well given the lack of a consumer. I am opening a pull request to avoid duplication of work should someone else need to use condition variables with spin locks.

### DIFF
--- a/include/sys/mutex.h
+++ b/include/sys/mutex.h
@@ -37,7 +37,8 @@ typedef enum {
 
 typedef struct {
 	struct mutex		m_mutex;
-	spinlock_t		m_lock;	/* used for serializing mutex_exit */
+	spinlock_t		m_lock;
+	kmutex_type_t		m_type;
 	kthread_t		*m_owner;
 } kmutex_t;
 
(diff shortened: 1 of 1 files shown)


tuxoko:
I don't like this. mutex and spinlock have complete different context. Hiding it behind same API is not a good idea. Unless you add mutex_spin_enter/mutex_spin_exit dedicated for spinlock. But in the end, I'd really like to a pure spinlock, because mutex is really big in compare.

behlendorf:
While this might be convenient I've resisted doing this for a long time since the size of a mutex is already too large.  Let's leave this as a last resort.  Plus if we actually wanted to emulate the proper illumos semantics for MUTEX_SPIN we'd need to use `spin_lock_irqsave()`, yuck.

behlendorf:
Closing, for the reasons mentioned above.  I'd prefer not to support this until there's a legitimate consumer which requires it.

PR #496, opened 2015-11-06
Fix taskq dynamic spawning

tuxoko:
Currently taskq_dispatch() will spawn new task with a condition that the caller
is also a member of the taskq. It's not clear why we need this condition and it
causes deadlock where a task on tq1 is waiting a task on tq2 to finish while the
task on tq2 wants to dispatch a task on tq1. So this patch removes it.

For example when you do:
zfs send pp/fs0@001 | zfs recv pp/fs0_copy

This will easily deadlock before this patch.

### DIFF
--- a/module/spl/spl-taskq.c
+++ b/module/spl/spl-taskq.c
@@ -538,7 +538,7 @@ taskq_cancel_id(taskq_t *tq, taskqid_t id)
 }
 EXPORT_SYMBOL(taskq_cancel_id);
 
-static int taskq_thread_spawn(taskq_t *tq, int seq_tasks);
+static int taskq_thread_spawn(taskq_t *tq);
 
 taskqid_t
 taskq_dispatch(taskq_t *tq, task_func_t func, void *arg, uint_t flags)
@@ -587,9 +587,8 @@ taskq_dispatch(taskq_t *tq, task_func_t func, void *arg, uint_t flags)
 	wake_up(&tq->tq_work_waitq);
 out:
 	/* Spawn additional taskq threads if required. */
-	if (tq->tq_nactive == tq->tq_nthreads &&
-	    taskq_member_impl(tq, current))
-		(void) taskq_thread_spawn(tq, spl_taskq_thread_sequential + 1);
+	if (tq->tq_nactive == tq->tq_nthreads)
+		(void) taskq_thread_spawn(tq);
 
 	spin_unlock_irqrestore(&tq->tq_lock, tq->tq_lock_flags);
 	return (rc);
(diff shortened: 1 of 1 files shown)


tuxoko:
The zfs send/recv is kind of like this:
```
zfs send -> dump_bytes --(wait)------------------->zio taskq -> pipe_wait --(wait)--->\\
->zfs recv -> txg_wait_synced --(wait)->\\           ^
->txg_sync -> zio_wait --(wait)----------------------+
```

So there's actually only one taskq involved. But it has circular dependency involving txg_sync thread

tuxoko:
@dweeezil 
>One of my lingering concerns about all this is what happens in one of these potentially deadlocking cases if the taskq in question doesn't have a sufficiently high tq_maxthreads. It would be interesting to identify these cases where concurrency is required in order to avoid a deadlock and determine whether there's a minimum requirement as to the number of threads required to avoid one.

Ideally, we should try to eliminate all circular dependency so that deadlock cannot occur no matter what the maxthread is. But it's like LOCKDEP issue, it's not going to be an easy taskq.

behlendorf:
> Ideally, we should try to eliminate all circular dependency so that deadlock cannot occur no matter
> what the maxthread is. But it's like LOCKDEP issue, it's not going to be an easy taskq.

Yes, LOCKDEP is a good analogy for this.  It's very challenging to identify all possible circular dependencies and just as hard to ensure no new ones are introduced.  However, this isn't a problem we need to solve in the context of this issue because both statically and dynamically sized taskqs will suffer the same deadlocks.  I think we can safely leave this issue for another time since it's clearly not currently a major issue.

Turning back to this issue it does look like we'll need to eliminate the `taskq_member_impl()` check to handle the more general case identified by @tuxoko.  That's a little unfortunate because we're now going to be more aggressively creating and destroying threads to handle an unlikely worst case scenario.  The `spl_taskq_thread_sequential` tuning was originally added to moderate that behavior but this change erodes that limiting somewhat.

How about extending this patch to also remove the `seq_taskq` argument from `taskq_thread_spawn()`.  That logic can be pulled out in to the `taskq_thread()` function to make it clear that's the only place it applies.  That would help simplify the code and make it more readable.

tuxoko:
@behlendorf 
Like this?

behlendorf:
@tuxoko yes, exactly what I had in mind.  @dweeezil are you OK with this change?

dweeezil:
@behlendorf Yes, the 5a7f551 patch looks just fine to me.  I like that it appears it will do the Right Thing if ```spl_taskq_thread_sequential``` is set to 0.

As you pointed out, if there are circular dependencies, the system can deadlock regardless of whether dynamic taskqs are being used.  I've been thinking of ways to write some debugging infrastructure to try to determine whether these dependencies actually happen.  Sort of our own taskq-specific lockdep thing, likely relying on wrappers to the various mutex and related functions.

behlendorf:
Merged as:

f5f2b87 Fix taskq dynamic spawning

PR #497, opened 2015-11-11
Remove kobj and vnode interfaces (WIP)

behlendorf:
__Only__ for review and feedback.  Removes the kobj, vnode, and spl_file, interfaces from the SPL.  A simplified implementation will be added to the ZFS source tree.   This existing code has served it's purpose well but it's needlessly complicated.  With minor changes on the ZFS side it can be vastly simplified.


ryao:
@behlendorf The changes you propose would require that I rewrite zfsonlinux/zfs#3907 to use Linux APIs at a time when I am preparing a refresh and have actually pushed it into production because the illumos developers prefer to receive patches with production code. This would make porting patches back and forth much more difficult. I would rather get the libzfs_core extensions finalized and sent to illumos before doing any major refactoring here.

behlendorf:
Closing as stale.  I'll refresh this and open a new PR when I can.

PR #501, opened 2015-11-20
Prevent rm modules.* when make install

tuxoko:
This was originally in e80cd06b8e0428f3ca2c62e4cb0e4ec54fda1d5c, but somehow
was changed and not working anymore. And it will cause the following error:

```
# modprobe zfs
modprobe: ERROR: ../libkmod/libkmod.c:506 lookup_builtin_file() could not open builtin file '/lib/modules/4.2.0-18-generic/modules.builtin.bin'
```


behlendorf:
Merged

d28c5c4 Prevent rm modules.* when make install

Issue #498, opened 2015-11-11
zfs/spl 0.6.5.3 build fails for linux-4.3 when configured as built-in modules


gedia:
It seems that building a 4.3 kernel with spl and zfs-kmod built-in fails. They build fine as modules. Linux-4.2.3 is not affected by this. I'm attaching a build log here as well as my kernel configuration.

GCC version is: gcc (Gentoo 5.2.0 p1.1, pie-0.6.4) 5.2.0
[build-log.txt](https://github.com/zfsonlinux/spl/files/32247/build-log.txt)
[config.txt](https://github.com/zfsonlinux/spl/files/32248/config.txt)

behlendorf:
Just taking a quick look it's not clear to me that it built any of the spl.  Since the errors are all about missing spl symbols that would make sense.

hunbalazs:
@behlendorf it didnt, it seems that (for some reason) it only builds ZFS but not SPL
```
# grep CONFIG_SPL .config
CONFIG_SPLIT_PTLOCK_CPUS=4
CONFIG_SPL=y
```
it works with kernel 4.2.6
(gcc version 4.9.3 (Gentoo 4.9.3 p1.1, pie-0.6.2))

sergiall:
It works with kernel 4.2.6 (gcc version 5.2.1 Debian stretch/sid,Ubuntu 15.10/16.04.
Both SPL & ZFS build ok and work fine,but for linux 4.3 it still fail for build :( .

behlendorf:
@sergiall @hunbalazs I've proposed two patches to fix this.  Any additional testing would be appreciated, I've verified the fix with v4.4-rc3 but it should fix all kernels.

hunbalazs:
Kernel built and running.
Thank you very much @behlendorf .

behlendorf:
@hunbalazs thanks, this has been merged to master and will appear in the next point release.

4e6f996 Fix --enable-linux-builtin

sergiall:
Wow! Thank you very much @behlendorf !

Issue #481, opened 2015-10-11
Kernel builtins broken in 0.6.5.2?


gcs-github:
Hello!

I stumbled upon an oddity today patching my kernel with SPL/ZFS 0.6.5.2.

Using `./copy-builtin` on a configured kernel tree works normally, and I'm then able to compile my kernel with SPL and then ZFS built-in without a hitch.

But, if after the `./copy-builtin` succeeds I issue a "make clean" or a "make distclean" in my kernel tree so I can distribute the patched kernel source or create different kernel configurations, in any subsequent kernel build attempt, SPL will fail to compile during the regular kernel build process, with the following message:

```
make[2]: *** No rule to make target 'spl/spl/spl-proc.o', needed by 'spl/spl/spl.o'.  Stop.
scripts/Makefile.build:403: recipe for target 'spl/spl' failed
make[1]: *** [spl/spl] Error 2
Makefile:949: recipe for target 'spl' failed
make: *** [spl] Error 2
```

This is with a vanilla Linux 4.2.3 source tree.

Linux kernel's "make clean" deleting more than it should, maybe?

behlendorf:
It certainly looks that way.  Thanks for filing this, we'll look in to it time permitting.

berney:
I had the same symptoms. In the Makefiles was a src= and obj= lines pointing to portage tmp dirs. I commented them out and then make would work.

behlendorf:
Merged:

4e6f996 Fix --enable-linux-builtin

PR #503, opened 2015-12-02
Fix --enable-linux-builtin

behlendorf:
Adding VPATH support, commit 37d7cd9, required that a `src`
and `obj` line be added to the top of the Makefiles.  They
must be removed from the Makefiles when builtin.

The code which adds the `spl/` directory to the top level
Makefile was failing due to the addition of the `certs/` path.
The search pattern has been adjusted to be more tolerant.

Issue #481
Issue #498


behlendorf:
4e6f996 Fix --enable-linux-builtin

PR #504, opened 2015-12-02
Use tsd to store tq for taskq_member

tuxoko:
To prevent taskq_member holding tq_lock and doing linear search, thus causing
contention. We store the taskq pointer to which the thread belongs in tsd.
This way taskq_member will not need to touch tq_lock, and tsd has per slot
spinlock. So the contention should be reduced greatly.

### DIFF
--- a/include/sys/taskq.h
+++ b/include/sys/taskq.h
@@ -124,7 +124,7 @@ extern void taskq_wait_id(taskq_t *, taskqid_t);
 extern void taskq_wait_outstanding(taskq_t *, taskqid_t);
 extern void taskq_wait(taskq_t *);
 extern int taskq_cancel_id(taskq_t *, taskqid_t);
-extern int taskq_member(taskq_t *, void *);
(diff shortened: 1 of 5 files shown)


ryao:
@tuxoko I have opened #505 for an alternative solution that should be more performant than TSD.

behlendorf:
Closing in favor of #505.

behlendorf:
Reopening due to issues exposed in the #505 patch.

behlendorf:
Replaced by #514.  @tuxoko I've taken your initial patch and refreshed it so it applies cleanly after the recent round of taskq improvements.  I also made sure it gets initialized before spl_kmem_cache_init() to prevent an ASSERT because a taskq is created by the kmem code and that now depends on this tsd code.

tuxoko:
@behlendorf Nice catch. But shouldn't ```spl_kmem_cache_init``` be after ```spl_taskq_init```?

Issue #500, opened 2015-11-15
locking in taskq_member() hurts performance on large-scale systems


dweeezil:
On large-scale systems with lots of CPU cores, the locking in ```taskq_member()``` becomes a significant contention point for zios dispatched to the single ```z_wr_iss``` taskq for heavy (high-frequency) IO workloads.  Under illumos, this operation is a single comparison operation whereas under ZoL, it involves taking a lock, iterating through a list and then releasing the lock.

This mainly affects the ```z_wr_iss``` taskq because it's the main zio taskq with a high volume of dispatches but for which only a single taskq is allocated.  Unfortunately, the obvious workaround of something like this:
```
diff --git a/module/zfs/spa.c b/module/zfs/spa.c
index 2e23a34..9480c2b 100644
--- a/module/zfs/spa.c
+++ b/module/zfs/spa.c
@@ -128,7 +128,7 @@ const zio_taskq_info_t zio_taskqs[ZIO_TYPES][ZIO_TASKQ_TYPES] = {
        /* ISSUE        ISSUE_HIGH      INTR            INTR_HIGH */
        { ZTI_ONE,      ZTI_NULL,       ZTI_ONE,        ZTI_NULL }, /* NULL */
        { ZTI_N(8),     ZTI_NULL,       ZTI_P(12, 8),   ZTI_NULL }, /* READ */
-       { ZTI_BATCH,    ZTI_N(5),       ZTI_P(12, 8),   ZTI_N(5) }, /* WRITE */
+       { ZTI_P(4, 10), ZTI_N(5),       ZTI_P(12, 8),   ZTI_N(5) }, /* WRITE */
        { ZTI_P(12, 8), ZTI_NULL,       ZTI_ONE,        ZTI_NULL }, /* FREE */
        { ZTI_ONE,      ZTI_NULL,       ZTI_ONE,        ZTI_NULL }, /* CLAIM */
        { ZTI_ONE,      ZTI_NULL,       ZTI_ONE,        ZTI_NULL }, /* IOCTL */
```
seems to expose other bottlenecks during high-throughput benchmarking and actually reduces performance.  I've not yet discovered why this happens but the above patch definitely reduces the impact of the locking in ```taskq_member()```.

behlendorf:
[...]

PR #505, opened 2015-12-03
Make taskq_member() use ->journal_info

ryao:
The ->journal_info pointer in the task_struct is reserved for use by
filesystems and because the kernel can have multiple file systems on the
same stack due to direct reclaim, each filesystem that touches
->journal_info in a callback function will save the value at the start
of its frame and restore it at the end of its frame.  This allows us to
safely use ->journal_info to store a pointer to the taskq's struct in
taskq threads so that ZFS code paths can detect the presence of a taskq.
This could break if the ZFS code were to use taskq_member from the
context of direct reclaim. However, there are no such uses of it in that
manner, so this is safe.

This eliminates an O(N) list traversal under a spinlock with an O(1)
unlocked pointer comparison.

Closes zfsonlinux/spl#500

### DIFF
--- a/include/linux/Makefile.am
+++ b/include/linux/Makefile.am
@@ -1,6 +1,7 @@
 COMMON_H =
 
 KERNEL_H = \
+	$(top_srcdir)/include/linux/api_compat.h \
 	$(top_srcdir)/include/linux/bitops_compat.h \
 	$(top_srcdir)/include/linux/compiler_compat.h \
 	$(top_srcdir)/include/linux/delay_compat.h \
--- a/include/linux/api_compat.h
+++ b/include/linux/api_compat.h
@@ -1,23 +1,46 @@
+typedef struct {
+	struct task_struct *lac_thread;
+	void *lac_journal_info;
+} api_cookie_t;
+
+static inline api_cookie_t
+spl_api_mark(void)
+{
+	api_cookie_t cookie;
+
+	cookie.lac_thread = current;
+	cookie.lac_journal_info = current->journal_info;
+	current->journal_info = NULL;
+
+	return (cookie);
+}
+
+static inline void
+spl_api_unmark(api_cookie_t cookie)
+{
+	ASSERT3P(cookie.lac_thread, ==, current);
+	ASSERT(current->journal_info == NULL);
+
+	current->journal_info = cookie.lac_journal_info;
+}
+
 #define LINUX_API_CALL_VOID(function, ...)			\
 do {								\
-	void *__journal_info = current->journal_info;		\
-	current->journal_info = NULL;				\
+	api_cookie_t cookie = spl_api_mark();		\
 	(void) (function)(__VA_ARGS__);				\
-	current->journal_info = __journal_info;			\
+	spl_api_unmark(cookie);					\
 } while(0)
 
 #define LINUX_API_CALL_TYPED(rc, type, function, ...)		\
 do {								\
-	void *__journal_info = current->journal_info;		\
-	current->journal_info = NULL;				\
+	api_cookie_t cookie = spl_api_mark();		\
 	(rc) = (type)(function)(__VA_ARGS__);			\
-	current->journal_info = __journal_info;			\
+	spl_api_unmark(cookie);					\
 } while(0)
 
 #define LINUX_API_CALL(rc, function, ...)			\
 do {								\
-	void *__journal_info = current->journal_info;		\
-	current->journal_info = NULL;				\
+	api_cookie_t cookie = spl_api_mark();		\
 	(rc) = (function)(__VA_ARGS__);				\
-	current->journal_info = __journal_info;			\
+	spl_api_unmark(cookie);					\
 } while(0)
--- a/include/linux/api_compat.h
+++ b/include/linux/api_compat.h
@@ -1,0 +1,23 @@
+#define LINUX_API_CALL_VOID(function, ...)			\
+do {								\
+	void *__journal_info = current->journal_info;		\
+	current->journal_info = NULL;				\
+	(void) (function)(__VA_ARGS__);				\
+	current->journal_info = __journal_info;			\
+} while(0)
+
+#define LINUX_API_CALL_TYPED(rc, type, function, ...)		\
+do {								\
+	void *__journal_info = current->journal_info;		\
+	current->journal_info = NULL;				\
+	(rc) = (type)(function)(__VA_ARGS__);			\
+	current->journal_info = __journal_info;			\
+} while(0)
+
+#define LINUX_API_CALL(rc, function, ...)			\
+do {								\
+	void *__journal_info = current->journal_info;		\
+	current->journal_info = NULL;				\
+	(rc) = (function)(__VA_ARGS__);				\
+	current->journal_info = __journal_info;			\
+} while(0)
--- a/include/sys/taskq.h
+++ b/include/sys/taskq.h
@@ -124,7 +124,7 @@ extern void taskq_wait_id(taskq_t *, taskqid_t);
 extern void taskq_wait_outstanding(taskq_t *, taskqid_t);
 extern void taskq_wait(taskq_t *);
 extern int taskq_cancel_id(taskq_t *, taskqid_t);
-extern int taskq_member(taskq_t *, void *);
+#define	taskq_member(taskq, thread)	((taskq) == ((thread)->journal_info))
 
 #define taskq_create_proc(name, nthreads, pri, min, max, proc, flags) \
     taskq_create(name, nthreads, pri, min, max, flags)
--- a/module/spl/spl-kmem-cache.c
+++ b/module/spl/spl-kmem-cache.c
@@ -27,6 +27,7 @@
 #include <sys/taskq.h>
 #include <sys/timer.h>
 #include <sys/vmem.h>
+#include <linux/api_compat.h>
 #include <linux/slab.h>
 #include <linux/swap.h>
 #include <linux/mm_compat.h>
@@ -199,7 +200,7 @@ kv_alloc(spl_kmem_cache_t *skc, int size, int flags)
 
 	if (skc->skc_flags & KMC_KMEM) {
 		ASSERT(ISP2(size));
-		ptr = (void *)__get_free_pages(lflags, get_order(size));
+		LINUX_API_CALL_TYPED(ptr, void *, __get_free_pages, lflags, get_order(size));
 	} else {
 		ptr = spl_vmalloc(size, lflags | __GFP_HIGHMEM, PAGE_KERNEL);
 	}
@@ -227,8 +228,9 @@ kv_free(spl_kmem_cache_t *skc, void *ptr, int size)
 
 	if (skc->skc_flags & KMC_KMEM) {
 		ASSERT(ISP2(size));
-		free_pages((unsigned long)ptr, get_order(size));
+		LINUX_API_CALL_VOID(free_pages, (unsigned long)ptr, get_order(size));
 	} else {
+		LINUX_API_CALL_VOID(vfree, ptr);
 		vfree(ptr);
 	}
 }
@@ -514,13 +516,13 @@ spl_emergency_alloc(spl_kmem_cache_t *skc, int flags, void **obj)
 	if (!empty)
 		return (-EEXIST);
 
-	ske = kmalloc(sizeof (*ske), lflags);
+	LINUX_API_CALL(ske, kmalloc, sizeof (*ske), lflags);
 	if (ske == NULL)
 		return (-ENOMEM);
 
-	ske->ske_obj = __get_free_pages(lflags, order);
+	LINUX_API_CALL(ske->ske_obj, __get_free_pages, lflags, order);
 	if (ske->ske_obj == 0) {
-		kfree(ske);
+		LINUX_API_CALL_VOID(kfree, ske);
 		return (-ENOMEM);
 	}
 
@@ -535,8 +537,8 @@ spl_emergency_alloc(spl_kmem_cache_t *skc, int flags, void **obj)
 	spin_unlock(&skc->skc_lock);
 
 	if (unlikely(!empty)) {
-		free_pages(ske->ske_obj, order);
-		kfree(ske);
+		LINUX_API_CALL_VOID(free_pages, ske->ske_obj, order);
+		LINUX_API_CALL_VOID(kfree, ske);
 		return (-EINVAL);
 	}
 
@@ -566,8 +568,8 @@ spl_emergency_free(spl_kmem_cache_t *skc, void *obj)
 	if (ske == NULL)
 		return (-ENOENT);
 
-	free_pages(ske->ske_obj, order);
-	kfree(ske);
+	LINUX_API_CALL_VOID(free_pages, ske->ske_obj, order);
+	LINUX_API_CALL_VOID(kfree, ske);
 
 	return (0);
 }
@@ -770,7 +772,7 @@ spl_magazine_alloc(spl_kmem_cache_t *skc, int cpu)
 	int size = sizeof (spl_kmem_magazine_t) +
 	    sizeof (void *) * skc->skc_mag_size;
 
-	skm = kmalloc_node(size, GFP_KERNEL, cpu_to_node(cpu));
+	LINUX_API_CALL(skm, kmalloc_node, size, GFP_KERNEL, cpu_to_node(cpu));
 	if (skm) {
 		skm->skm_magic = SKM_MAGIC;
 		skm->skm_avail = 0;
@@ -792,7 +794,7 @@ spl_magazine_free(spl_kmem_magazine_t *skm)
 {
 	ASSERT(skm->skm_magic == SKM_MAGIC);
 	ASSERT(skm->skm_avail == 0);
-	kfree(skm);
+	LINUX_API_CALL_VOID(kfree, skm);
 }
 
 /*
@@ -806,7 +808,7 @@ spl_magazine_create(spl_kmem_cache_t *skc)
 	if (skc->skc_flags & KMC_NOMAGAZINE)
 		return (0);
 
-	skc->skc_mag = kzalloc(sizeof (spl_kmem_magazine_t *) *
+	LINUX_API_CALL(skc->skc_mag, kzalloc, sizeof (spl_kmem_magazine_t *) *
 	    num_possible_cpus(), kmem_flags_convert(KM_SLEEP));
 	skc->skc_mag_size = spl_magazine_size(skc);
 	skc->skc_mag_refill = (skc->skc_mag_size + 1) / 2;
@@ -817,7 +819,7 @@ spl_magazine_create(spl_kmem_cache_t *skc)
 			for (i--; i >= 0; i--)
 				spl_magazine_free(skc->skc_mag[i]);
 
-			kfree(skc->skc_mag);
+			LINUX_API_CALL_VOID(kfree, skc->skc_mag);
 			return (-ENOMEM);
 		}
 	}
@@ -843,7 +845,7 @@ spl_magazine_destroy(spl_kmem_cache_t *skc)
 		spl_magazine_free(skm);
 	}
 
-	kfree(skc->skc_mag);
+	LINUX_API_CALL_VOID(kfree, skc->skc_mag);
 }
 
 /*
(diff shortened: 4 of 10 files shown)


dweeezil:
@ryao Nice!  When I first discovered this issue, I dug through ```task_struct``` looking for a place to stash the taskq but there are so many members in the struct that I didn't really know where to start.  This, combined with the dispatch elimination patch on 16K stack systems should help a lot in the zio execute path.

ryao:
@dweeezil I am glad someone is working on profiling zio execute. Be sure to keep me in the loop. :)

ryao:
@dweeezil If we were to need more pointers in the task_struct for the taskq, I imagine that the pointers used for managing userspace things are candidates for use as long as we restrict their use to kernel threads such as the taskq threads. Specifically, `->real_parent`, `->parent`, `->vfork_done`, `->last_siginfo`, and `->splice_pipe`. However, the manner in which they are used by the kernel would need to be checked.

What appealed to me about `->journal_info` is that it has been around a long time, has consumers that would never clobber what we put there unless we use taskq_member from a callback context (which we don't) and is intended for filesystem usage.

In either case, this approach has no false positives, because the pointer will naturally point to a different object at a different memory address when in use, even if taskq_member is called from a userland thread when something is actually using whatever pointer we decided to use. The only danger is calling taskq_member from a callback context in an actual taskq thread, where we would get a false positive. Thankfully, we do not do that and likely never will. :)

tuxoko:
> this approach has no false positives, because the pointer will naturally point to a different object at a different memory address when in use

While I agree it is not likely to have false positive, but there's no guarantee that ```journal_info``` will be only used as pointers.

> The only danger is calling taskq_member from a callback context in an actual taskq thread, where we would get a false positive.

What do you mean by that? Can you give an example?

@ryao 
The idea of using a member in task_struct is sure tempting. But I want to make sure, have you test the cases where taskq thread go into other filesystem? For example, vdev_file, zfs send > file, etc.

behlendorf:
@ryao this sure is a tempting solution.  If we could safely use even a single pointer in the task struct could potentially solve several issues.  For example, we have this exact same problem in the tsd implementation, specifically it would be helpful if we could hang some data off the task_struct.  But since there was no obvious safe way to do this the tsd uses a hash.

What about instead of using `journal_info` for the taskq member point we uses it to optimize the tsd implementation.  Then we'd be able to easily add more than just a single value to the task_struct.

ryao:
@tuxoko I checked every use of ->journal_info in the kernel. It is only used for pointers. The case where a taskq thread goes into another file system works fine because the other filesystem will:

1. Never callback into a ZFS function that uses `taskq_member()`
2. Save the value of `->journal_info` before changing it and restore it before it returns.

As for an example of the danger, one doesn't actually exist in the wild, so it is hard to give an actual example. However, hypothetically speaking, you could imagine a stack that goes taskq_thread -> ext4 -> arc shrinker -> `taskq_member()`, which would produce a false positive.

@behlendorf We could try using one of the pointers used by userspace for `taskq_member()` and use `->journal_info` for TSD, but so far, I have only verified that ->journal_info is safe to use. One of us will need to check the other candidates to see if they work here. At the moment, I will not be able to look at them until I finish the refresh of zfsonlinux/zfs#3907.

If we use `->journal_info` for TSD, we will need to take care to save the current value and then restore it after exiting from ZFS like every other filesystem does when using `->journal_info`. Otherwise, we could cause data corruption in other filesystems when we clobber their `->journal_info` values.

behlendorf:
I just wanted to toss out the idea of using it for tsd.  If @dweeezil and @tuxoko sign-off on using `->journal_info` for the taskq thread that's OK with me.  It has a nice simplicity to it and it can always be changed as needed.

> For the curious, I ported the Illumos taskq code last year, but didn't put it in a PR due to it using mutexes in an interrupt context that required a double dispatch that @behlendorf told me he did not like when I mentioned it to him.

It would be great if you could still put it in a pull request, then we could get more feedback.  As I've said before I'm not opposed to using the illumos taskq code.  But before we throw out the working stable implementation somebody needs to definitively show the illumos code is actually _better_.

ryao:
@behlendorf I will try to do that after I refresh the PR that my boss wants me to focus on pushing to get merged. The illumos taskq code is really cool and now that I think of it, we could probably solve the double dispatch issue by switching it to use spin locks. I have an implementation of MUTEX_SPIN in a local branch somewhere that I could dig out for that. :)

ryao:
@behlendorf It occurs to me that if we are careful about how we manage things (and maybe add some magic numbers for added caution), we could probably use `->journal_info` in both TSD and taskq_member. When we enter a code path that uses TSD, we could save the current pointer and restore it upon leaving, just like any other consumer would. We just need to be sure we do not not have `taskq_member()` in a TSD "critical section" and we would be okay. That is where some magic numbers would come in handy.

tuxoko:
The problem with tsd is that there's an insane use case in ioctl that it would survive across syscalls into userspace. Unless we eliminate it, we should not put tsd into task_struct.

ryao:
@tuxoko Good point

ryao:
@behlendorf @tuxoko On second thought, we could do TSD using a pointer in the `task_struct`, provided that we continue to use the hash table on entry into the ZoL code to see if we should initialize the pointer and remove it on exit. This means that we would only need to do the lookup once on entry, but I am not sure if that would be worthwhile given how little TSD is used.

@tuxoko @dweeezil Are you alright with merging this patch?

behlendorf:
Unless we can eliminate the hash entirely, and it doesn't sound like we can, I don't think it's worth pursuing the tsd idea.

ryao:
@tuxoko @dweeezil I think @behlendorf is waiting for you to ACK/NACK this before merging it.

tuxoko:
@behlendorf @ryao I'm ok with this.

dweeezil:
@behlendorf @ryao LGTM.  I researched the use of ```journal_info``` over time and it seems to be safe to use in this way for this purpose.

behlendorf:
I agree this looks good.  Thanks,  I'll add the needed signed-off-by's and merge it.

ryao:
Some people I met on IRC did some benchmarks of deduplication performance last night. Unfortunately, the test setup was somewhat odd. They were doing vdbench over the network (I think they used iSCSI) on zvols served by ZFS on RAID5 on SSDs. They had 128GB of RAM and a 12-core Xeon E5-2695 v2 with HT disabled. Their benchmark numbers were fairly variable. However, in general, this patch helped IOPS and reduced response times. They were seeing IOPS spikes to 50,000 when previously 45,000 was the best and average response time was trimmed the range from 3ms to 5ms by about 0.8ms. Prior to this patch, ZoL was having trouble saturating the 12-core Intel Xeon CPU (HT was disabled) before the patch and 3.5% of time spent was spent in `taskq_member()`. ZoL was saturating the CPU.

Also, the system was seeing maximum IO spikes to >30 seconds before switching from 0.6.5.3 to HEAD. After switching, those disappeared. Worst case IO was something like 1.3 seconds, but that is incredibly rare. My expectation is that the large stacks patch is the cause as we are doing less dispatching.

dweeezil:
I'll try running a few benchmarks today.  I think it's pretty clear this will help the "lots of small writes" case, especially with the 16K stack optimization patch, mainly due to the geometry of the write issue taskq (single queue with large max threads).

behlendorf:
@ryao thanks for sharing those very encouraging results!  Merged as:

a430c11 Make taskq_member() use ->journal_info

ryao:
@behlendorf Anytime. :)

behlendorf:
Reverted a430c11, this commit results in the following BUG on at least 4.1 kernels.  Fast or not this is going to need to be revisited.  If you recently submitted pull requests against ZFS you may want to force update them or you'll get false failures from this.

```
[ 1081.186286] ------------[ cut here ]------------
[ 1081.188273] kernel BUG at fs/jbd2/transaction.c:425!
[ 1081.188273] kernel BUG at fs/jbd2/transaction.c:425!
[ 1081.190167] invalid opcode: 0000 [#1] SMP 
[ 1081.190167] Modules linked in: zfs(POE) zcommon(POE) zunicode(POE) znvpair(POE) zavl(POE) splat(OE) spl(OE) ext2(E) loop(E) ipv6(E) binfmt_misc(E) evbug(E) evdev(E) psmouse(E) i2c_piix4(E) i2c_core(E) acpi_cpufreq(E) ixgbevf(E) button(E) ext4(E) crc16(E) jbd2(E) mbcache(E) ata_piix(E) libata(E) scsi_mod(E) dm_mirror(E) dm_region_hash(E) dm_log(E) dm_mod(E) [last unloaded: spl]
[ 1081.190167] CPU: 3 PID: 26048 Comm: spl_system_task Tainted: P           OE   4.1.10-17.31.amzn1.x86_64 #1
[ 1081.190167] Hardware name: Xen HVM domU, BIOS 4.2.amazon 05/06/2015
[ 1081.190167] task: ffff8803f9c31980 ti: ffff880351ee8000 task.ti: ffff880351ee8000
[ 1081.190167] RIP: 0010:[<ffffffffa00eaaee>]  [<ffffffffa00eaaee>] jbd2__journal_start+0x1de/0x1e0 [jbd2]
[ 1081.190167] RSP: 0018:ffff880351eeba18  EFLAGS: 00010212
[ 1081.190167] RAX: ffff8803fc91be00 RBX: ffff880037ac0000 RCX: 0000000000000050
[ 1081.190167] RDX: ffff880314c214c2 RSI: 0000000000000002 RDI: ffff880037ac0000
[ 1081.190167] RBP: ffff880351eeba58 R08: 0000000000000001 R09: 00000000000013eb
[ 1081.190167] R10: ffff880385e33f00 R11: 0000000000000000 R12: 0000000000000002
[ 1081.190167] R13: 0000000000000000 R14: ffffffffa0123332 R15: ffff880351eebcf0
[ 1081.190167] FS:  0000000000000000(0000) GS:ffff88040fc60000(0000) knlGS:0000000000000000
[ 1081.190167] CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
[ 1081.190167] CR2: 00007fbbeb1cb978 CR3: 0000000001a0e000 CR4: 00000000001406e0
[ 1081.190167] Stack:
[ 1081.190167]  ffff880351eebaf8 ffffffff810a30c3 ffff880351eeba78 ffff8803fa777000
[ 1081.190167]  0000000000000002 0000000000000000 ffffffffa0123332 ffff880351eebcf0
[ 1081.190167]  ffff880351eebaa8 ffffffffa014d129 ffff880351eeba78 000013eb00000001
[ 1081.190167] Call Trace:
[ 1081.190167]  [<ffffffff810a30c3>] ? update_sd_lb_stats+0x123/0x550
[ 1081.190167]  [<ffffffffa0123332>] ? ext4_dirty_inode+0x32/0x70 [ext4]
[ 1081.190167]  [<ffffffffa014d129>] __ext4_journal_start_sb+0x69/0xe0 [ext4]
[ 1081.190167]  [<ffffffffa0123332>] ext4_dirty_inode+0x32/0x70 [ext4]
[ 1081.190167]  [<ffffffff811f451f>] __mark_inode_dirty+0x15f/0x3b0
[ 1081.190167]  [<ffffffff811e3c07>] generic_update_time+0x87/0xe0
[ 1081.190167]  [<ffffffff811e3e2a>] file_update_time+0x9a/0xf0
[ 1081.190167]  [<ffffffff8115b730>] __generic_file_write_iter+0x80/0x1d0
[ 1081.190167]  [<ffffffff8115a44c>] ? pagecache_get_page+0x2c/0x1d0
[ 1081.190167]  [<ffffffffa0116d91>] ext4_file_write_iter+0x101/0x390 [ext4]
[ 1081.190167]  [<ffffffff811c8f11>] __vfs_write+0xb1/0xf0
[ 1081.190167]  [<ffffffff811c9579>] vfs_write+0xa9/0x1b0
[ 1081.190167]  [<ffffffffa0794024>] ? spa_taskq_dispatch_ent+0x74/0xa0 [zfs]
[ 1081.190167]  [<ffffffffa0297da8>] vn_rdwr+0x68/0xe0 [spl]
[ 1081.190167]  [<ffffffffa07ad3de>] vdev_file_io_strategy+0x6e/0xa0 [zfs]
[ 1081.190167]  [<ffffffffa0296ee5>] taskq_thread+0x215/0x440 [spl]
[ 1081.190167]  [<ffffffff81094bc0>] ? wake_up_state+0x20/0x20
[ 1081.190167]  [<ffffffffa0296cd0>] ? taskq_cancel_id+0x120/0x120 [spl]
[ 1081.190167]  [<ffffffff81087359>] kthread+0xc9/0xe0
[ 1081.190167]  [<ffffffff81087290>] ? kthread_create_on_node+0x180/0x180
[ 1081.190167]  [<ffffffff814e18a2>] ret_from_fork+0x42/0x70
[ 1081.190167]  [<ffffffff81087290>] ? kthread_create_on_node+0x180/0x180
```

ryao:
@behlendorf Thanks for the backtrace. I am looking into it.

ryao:
I misread how ext4 touched `->journal_info`. It is not doing save/restore like I had thought. We can:

1. Try to do save/restore for it in vn_rdwr.
2. Try to use a different pointer.
3. Try something else.

Direct reclaim could mess us up with 1 because of ext4's behavior, but 2 might be okay with `->last_siginfo`. I need to check what stops userland from ptracing a kernel thread.

dweeezil:
Apparently I was also too quick to judge the use of ```journal_info``` as safe in the other filesystems.  I looked around a bit more and found this interesting comment in lustre which may be relevant: https://github.com/torvalds/linux/blob/v4.3/drivers/staging/lustre/lustre/obdclass/cl_object.c#L506.

As to doing save/restore in ```vn_rdwr```, that would fix zfs->ext4 but what about ext4->zfs stacking?

In light of this, it's not clear to me why ext4->btrfs or btrfs->ext4 stacking would be safe, either.

Unless there's another ```task_struct``` member available, tsd might be the best solution after all.

ryao:
@dweeezil ext4 -> zfs stacking is not a problem because the taskq threads don't stack on other filesystems.

As for why btrfs, ext4 and others can share `->journal_info` safely, it might be that the usage was designed around the assumption that you will never see one filesystem calling into another during a journal transaction (i.e. use of `->journal_info`) outside of direct reclaim, which deals with dirty writeback and hopefully can avoid using `->journal_info` entirely. That is not the case in the taskq threads.

@behlendorf If you could provide instructions to reproduce (although I suspect that I can guess), I should be able to try some ideas. This would be the second time this year that I recall a performance improvement required a workaround for a bad interaction between our code changes and a call into kernel code touched the task_struct. The first being this:

https://github.com/zfsonlinux/zfs/commit/37f9dac592bf5889c3efb305c48ac39b4c7dd140#diff-7cdd0a8aff43697fe0594301b843006aR500

behlendorf:
> 1) Try to do save/restore for it in vn_rdwr.
> 2) Try to use a different pointer.

While both of these may be technically possible I don't think they'd be a wise choice.  We need to make too many assumptions about the kernel, and while those assumptions may be true today they can change at any time.  The resulting implementation has the potential to be brittle.

> 3) Try something else.

While not quite as efficient as this `journal_info` solution @tuxoko's tsd based patch in #504 is far faster than the existing implementation.  And because it's based on the stable tsd interface we don't have to worry about upstream kernel changes accidentally breaking it in a subtle way.

My preference would be to adopt #504.

@ryao the failure was [exposed by the buildbot](http://build.zfsonlinux.org/builders/Amazon%202015.09%20x86_64%20%28TEST%29/builds/85/steps/shell_8/logs/console).  It's something we could have caught before the merge but as you know for the moment the SPL matches are largely only manually tested before being merged.  So occasionally something gets missed.

ryao:
@behlendorf I have pushed an initial stab at resolving the issue, although I wrote it before seeing your reply. I realize that this push missed a few possible points where save/restore is needed (e.g. `kthread_create()` and `spl_kmem_cache_create()`/`spl_kmem_cache_destroy()`, but I am pushing it now for your thoughts.

behlendorf:
I think one of the most appealing things about your original patch what its simplicity.  It was absolutely clear how it worked and it would be easy to maintain.  I think you'll agree that this updated patch, while it looks like it should work, is far more subtle and it would be easy for someone to accidentally break.

Because it will be maintainable long term I'd prefer the tsd solution.  But I'd like to hear @dweeezil's and @tuxoko's throughts on this before we decide anything.

ryao:
@behlendorf The TSD code will also suffer contention issues due to the global spinlock, although it will not be quite as bad. With this approach, the use of a spinlock in `taskq_member` is gone and spinlock usage remaining outside of opportunistic spin is negligible. I would prefer to explore the idea of using a pointer more as the pointer is superior. My two ideas for doing that are either wrapping the callbacks with save/restore or picking another pointer. Given that filesystems are supposed to share the `->journal_info` pointer, I prefer the latter.

I have revised the second patch to improve maintainability and  increase the number of wrapped functions. There are only a finite number of invocations of problematic kernel functions and I believe that wrapping all of them is doable. In specific, all `__get_free_pages`, `free_pages`, `kfree`, `kmalloc`, `kzalloc`, `vfree`, `vmalloc` and `vfs_*` invocations in the spl module have been switched to use the new wrappers while `kthread_create()` had to be manually wrapped due to it being a macro. I have not touched the splat code yet though.

ryao:
It probably is worth mentioning that if we use statement expressions, we can avoid most of these modifications entirely by defining macros for those functions and using autotools checks to verify that the macros are working properly.

https://gcc.gnu.org/onlinedocs/gcc/Statement-Exprs.html#Statement-Exprs

That GNU C extension makes things harder to understand and is likely non-portable, so I did not use it.

behlendorf:
> the TSD code will still suffer contention issues due to the global spinlock

Contention should be minimal because with the TSD code we're only locking one of 512 the hash buckets each of which contains a spin lock to do the lookup.  This could be increased further for very little cost if contention is an issue.  We are not serializing everything through one lock.

I agree that using the task_struct would be more efficient but doing that in a safe, understandable, and maintainable way is a challenge.  Feel free to investigate further, but in my mind whatever solution we end up with should ideally be 1) easily understandable, 2) largely contained to the taskq code, 3) maintainable.  It does not necessarily have to be strictly the fastest.

tuxoko:
@ryao 
It's not only the spl API, any call to kernel API might also have allocation, for example bio_alloc is the more obvious one. There might be other not so obvious, and it's too difficult to audit every API we use.

ryao:
@behlendorf If it performs well, lets merge it then. I really do like the idea of making this lock-free/atomic-free though.

@tuxoko That is why I defined the header.

ryao:
As an addendum, we could preemptively wrap `ilookup` and `igrab` to handle contexts such as `ZFS -> ext4 -> ZFS file backed pool` in addition to the bio_* functions in vdev_disk.c. I do not see much else, but I am still looking.

ryao:
@behlendorf These wrappers would also make it fairly easy to insert dedicated trace points for profiling our usage of the kernel API, which is a pain to do right now. At the moment, how do you filter stacks that are only invocations from our code and how performant is that?

I suspect that these wrappers are good idea even if we go with the TSD solution.

ryao:
@behlendorf These wrappers also should make it fairly easy to insert dedicated trace points for profiling our usage of the kernel API, which is a pain to do right now. At the moment, how do you filter stacks that are only invocations from our code and how performant is that?

I suspect that these wrappers are good idea even if we go with the TSD solution. In that case, leveraging them for making `->journal_info()` safe could still be considered a simpler patch.

dweeezil:
I'll run another round of tests today comparing the tsd patch against the new journal_info wrapper patch.  Either will likely be a huge improvement over the old implementation.

As to statement expressions, it appears they're already used in SPL in a few places and IIRC, they're used in ABD in ZFS as well.  And of course, the kernel uses them a lot.

dweeezil:
I'm testing with fio:
```
[test]
        blocksize=4k
        scramble_buffers=1      
        disk_util=0
        invalidate=0
        size=1g
        numjobs=32
        create_serialize=1
        direct=0
        directory=/tank/fish  # with recordsize=4k
        group_reporting=1
        ioengine=pvsync
        rw=randwrite
        time_based=1
        runtime=5m
        fsync=0
```
on a pool with 20 top-level whole-disk vdevs to get a decent amount of possible IOPS.  The 32 test files are pre-cached by fully reading them after importing the pool.  The spl/zfs build is currently using ```--enable-debug``` but ```zfs_flags=0``` is set to lessen its impact.

Initial tests to gauge the effect of the 16-k stack patch, which should reduce the calls to ```taskq_member``` and eliminate extra async dispatching, are very surprising.  With the 16-k stack patch (current master code), it gets 80328 iops at 321MiB/s.  Reverting the 16-k stack patch causes it to get 96824 iops at 387MiB/s.  I'm not quite sure how to explain this yet and am wondering if the fact that the test system has 80 hardware threads available could have something to do with it.

I'm posting these number now because I was surprised by them.  I'm going to switch to non-debug builds before continuing further (and trying the tsd patch).

ryao:
@dweeezil Thanks for the initial numbers. They are certainly surprising.

ryao:
@dweeezil I suggest using these benchmarks:

https://gist.github.com/brendangregg/7270ff9698c70d9e7496

behlendorf:
@dweeezil that's not what I expected, but it strongly suggests that for at least some configurations and workloads it's actually beneficial to allow the sync thread to hand off work to other threads.

dweeezil:
@ryao Thanks for the link, those are useful benchmarks.  I chose a write-only benchmark to stress the single write issue taskq.  I'll run some different tests next go round.

behlendorf:
Closing in favor of #514.

PR #480, opened 2015-10-07
add spin_lock_irqsave_nolockdep and mutex type MUTEX_NOLOCKDEP

ofaaland:
When running a kernel with CONFIG_LOCKDEP=y, lockdep reports possible
recursive locking in some cases and possible circular locking dependency
in others, within the SPL and ZFS modules.

When lockdep detects these conditions, it disables further lock analysis
for all locks.  This causes /proc/lock_stats not to reflect full
information about lock contention, even in locks without dependency
issues.

This commit creates new macros spin_lock_irqsave_nolockdep and
spin_unlock_irqrestore_nolockdep.  These macros use lockdep_off() and
lockdep_on() to temporarily disable lockdep when taking/releasing the
locks.

This commit also creates a new type of mutex, MUTEX_NOLOCKDEP.  This
mutex type causes subsequent attempts to take or release those locks to
be wrapped in lockdep_off() and lockdep_on().

At this time it's unknown whether this method distorts the statistics
reported by lock_stats.

This commit uses the new spin_lock macros for one such offending lock.

See also commit  "Identify locks flagged by lockdep" at https://github.com/zfsonlinux/zfs/pull/3895

Combined, these two commits mark all offending locks at this time, for
later analysis and fixes or annotation.

PR #506, opened 2015-12-03
Don't use tq->tq_lock_flags

tuxoko:
The flags argument in spin_lock_irqsave is modified out side of spin_lock
context. We cannot use a shared variable like tq->tq_lock_flags for them. This
patch removes it and uses local variable for the flags.

### DIFF
--- a/include/sys/taskq.h
+++ b/include/sys/taskq.h
@@ -60,7 +60,6 @@ typedef void (task_func_t)(void *);
 
 typedef struct taskq {
 	spinlock_t		tq_lock;       /* protects taskq_t */
-	unsigned long		tq_lock_flags; /* interrupt state */
 	char			*tq_name;      /* taskq name */
 	struct list_head	tq_thread_list;/* list of all threads */
 	struct list_head	tq_active_list;/* list of active threads */
--- a/module/spl/spl-taskq.c
+++ b/module/spl/spl-taskq.c
@@ -71,7 +71,7 @@ task_km_flags(uint_t flags)
  * is not attached to the free, work, or pending taskq lists.
  */
 static taskq_ent_t *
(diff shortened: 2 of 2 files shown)


ryao:
Nice find!

ryao:
Could you provide an example of something that modifies the flags argument?

That said, I like the idea of using a local variable here.

behlendorf:
Thanks!  Using a local variable here is definitely the right thing to do.  That said did you actually encounter a problem because of the shared variable or is this just cleanup?

tuxoko:
I don't really have a real life example of the problem.
But the EFLAGS register contains so many things, I can only imagine the horrible things that may happen if you restore it from another CPU's flags.

One possible example in current implementation is the dynamic_taskq. Because the dynamic_taskq->tq_lock is recursively locked in the taskq_dispatch path. This could override the flags saved by its thread with an irq disabled flags. Causing the thread stays in irq disable indefinitely.

dweeezil:
@tuxoko This is very interesting sounding.  I've got an unexplained hang I can reproduce with my taskq kstat patch (I had already moved its use of ```tq->tq_lock_flags``` to a local variable before seeing this pull request). I'll post a few more details in the pull request.

ryao:
@tuxoko I was initially skeptical of this, but you are correct:

http://lxr.free-electrons.com/source/include/linux/spinlock_api_smp.h#L106

The caller will save its irq flags to the flags variable *before* grabbing the spin lock. This means we are clobbering the irq mask in this code. This potentially explains some rare issues that I have seen in the taskq code.

behlendorf:
I've reconciled this patch with the lockdep work in #480 and have merged it to master.

066b89e Don't use tq->tq_lock_flags

PR #507, opened 2015-12-08
skip module compilation test when cross-compiling

kdomanski:


behlendorf:
@kdomanski when cross-compiling we can't disable these tests because they're used to determine what interfaces are provided by the host kernel.  What needs to happen here is we need to verify that this simple module can be successfully cross-compiled.

Exactly how are you trying to cross-compile and why is this step failing?

kdomanski:
I'm building SPL with Portage for a custom CoreOS image. This particular test fails with
`configure: error: cannot run test program while cross compiling`. So this particular module can never be cross-compiled anyway, unless you fool `configure` into thinking you're not cross-compiling.

behlendorf:
I think part of the confusion here is that there are two checks going on in this macro.  There's the first SPL_LINUX_TRY_COMPILE check which determines if modules can be built, and there's a second AC_RUN_IFELSE license compatibility check which is what's failing.  This patch should preserve the first check and disable the second as you've done when cross compiling.

tuxoko:
Can't the second one just use host build?

behlendorf:
@tuxoko good point.  It absolutely could and that would be better.

kdomanski:
I don't know how to force a host build for the second test, but I simply made it skip on cross-build.

behlendorf:
Nice concise fix.  I thought for a moment you didn't push the full fix but after reading the [documentation](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/html_node/Runtime.html) I see this is all that's needed.  This LGTM, but just to be completely clear and to document this exactly what command are you running to cross-compile.

behlendorf:
I see @tuxoko opened an alternate fix in #509 which doesn't rely on being able to execute the code.  @kdomanski since you're already setup to test this it would be great if you could validate this approach.

kdomanski:
@behlendorf I use gentoo portage with [this ebuild](https://github.com/gentoo/gentoo/blob/master/sys-kernel/spl/spl-0.6.5.3.ebuild) on CoreOS SDK.

@tuxoko's fix works for me.

behlendorf:
I think we should go with @kdomanski patch since it will only impact the cross-compile case.  Using BUILD_BUG_ON would be nice but if we can't count on it's reliability on all platforms and gcc versions we'll get incorrect results.  @kdomanski @tuxoko if this sounds reasonable to you I'll merge the patch in this pull request.

tuxoko:
Yeah, I'm fine with this.

behlendorf:
Reopening and reverting the patch.  I just discovered after merging this change that it doesn't work with autoconf 2.63.  We'll need to do something else.

kdomanski:
Pushed a little less elegant approach. Please test.

behlendorf:
Let's try this again, merged as: 

e0ed96f Skip GPL-only symbols test when cross-compiling

PR #509, opened 2015-12-10
Don't use AC_RUN_IFELSE when testing license

tuxoko:
AC_RUN_IFELSE would cause cross-compiling failed, so we changed the license
testing to use BUILD_BUG_ON and rely on compiler to do constant folding so we
don't need to run anything.


tuxoko:
It works on my desktop, but not on buildbot. Probably older gcc can't do that.

kdomanski:
Works for me.

PR #510, opened 2015-12-11
fix build issue on some configured Kernels.

zgock:
spl fails to build with some "Configured" kernels (ex. openSUSE xen Kernel)
this change should make same binaries with C compiler optimization.
thx.


behlendorf:
I agree, this is much nicer approach since it allows us to use the ZFS code unmodified.  Let me verify it doesn't cause issues on any other platforms.  Also we'll want to remove the `#define` entirely not just comment it out.

behlendorf:
LGTM, merged as:

70d80b3 Fix build issue on some configured kernels

PR #515, opened 2015-12-18
Use spl_fstrans_mark instead of memalloc_noio_save

tuxoko:
For earlier version of kernel with memalloc_noio_save, it only turns off
__GFP_IO but leaves __GFP_FS untouched during direct reclaim. This would cause
threads to direct reclaim into ZFS and cause deadlock.

Instead, we should stick to using spl_fstrans_mark. Since we would explicitly
turn off both __GFP_IO and __GFP_FS before allocation, it would work on every
version of kernel.

### DIFF
--- a/include/sys/vmem.h
+++ b/include/sys/vmem.h
@@ -36,7 +36,6 @@ extern vmem_t *zio_alloc_arena;
 extern vmem_t *zio_arena;
 
 extern size_t vmem_size(vmem_t *vmp, int typemask);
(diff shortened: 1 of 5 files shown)


dweeezil:
Nice find!  It appears this would impact kernels 3.9-3.17 which might be worth mentioning and also to reference torvalds/linux@934f307.

behlendorf:
LGTM, I like the idea of depending on fewer interfaces.

PR #519, opened 2016-01-07
Don't hold mutex until release cv in cv_wait

tuxoko:
If a thread is holding mutex when doing cv_destroy, it might end up waiting a
thread in cv_wait. The waiter would wake up trying to aquire the same mutex
and cause deadlock.

We solve this by move the mutex_enter to the bottom of cv_wait, so that
the waiter will release the cv first, allowing cv_destroy to succeed and have
a chance to free the mutex.

This would create race condition on the cv_mutex. We use xchg to set and check
it to ensure we won't be harmed by the race. This would result in the cv_mutex
debugging becomes best-effort.

Also, the change reveals a race, which was unlikely before, where we call
mutex_destroy while test threads are still holding the mutex. We use
kthread_stop to make sure the threads are exit before mutex_destroy.

### DIFF
--- a/module/spl/spl-condvar.c
+++ b/module/spl/spl-condvar.c
@@ -80,6 +80,7 @@ static void
 cv_wait_common(kcondvar_t *cvp, kmutex_t *mp, int state, int io)
 {
 	DEFINE_WAIT(wait);
+	kmutex_t *m;
 
 	ASSERT(cvp);
 	ASSERT(mp);
@@ -87,11 +88,11 @@ cv_wait_common(kcondvar_t *cvp, kmutex_t *mp, int state, int io)
 	ASSERT(mutex_owned(mp));
 	atomic_inc(&cvp->cv_refs);
 
-	if (cvp->cv_mutex == NULL)
-		cvp->cv_mutex = mp;
-
+	m = ACCESS_ONCE(cvp->cv_mutex);
+	if (!m)
+		m = xchg(&cvp->cv_mutex, mp);
 	/* Ensure the same mutex is used by all callers */
-	ASSERT(cvp->cv_mutex == mp);
+	ASSERT(m == NULL || m == mp);
 
 	prepare_to_wait_exclusive(&cvp->cv_event, &wait, state);
 	atomic_inc(&cvp->cv_waiters);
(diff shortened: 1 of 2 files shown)


tuxoko:
Fix zfsonlinux/zfs#4166 and second part of zfsonlinux/zfs#4106

dweeezil:
I will try my existing tests with this.

tuxoko:
Update: Change `ASSERT` to remove `if` and modify __cv_timedwait_hires.

behlendorf:
@tuxoko  nice!  This isn't nearly as disruptive as I'd feared.  The ref counts nicely ensure the memory remains valid while cv_wait (and friends) finish up.  Moving the mutex to the end of the function removes the lock inversion.  We've always relied on the caller not destroying the mutex prematurely so there no new concern there.  That said, this kind of thing can be subtle so we'll definitely want to stress the new code... and it sounds like @dweeezil is already on that.  Awesome, thanks guys!

dweeezil:
Looks good, see https://github.com/zfsonlinux/zfs/issues/4106#issuecomment-169828889.

tuxoko:
The splat complains a lot in https://github.com/zfsonlinux/zfs/pull/4173
I'll need to look into it.

dweeezil:
@tuxoko What are your thoughts on this so far?  I didn't look at the buildbot errors but I can easily get an assert from condvar:broadcast1 in debug builds because mutex is not held.  Is this what the bots are triggering?  So far, I've not seen the assert in testing with zfs, however, it would seem that ```zio_wait()``` might be able to trigger it and also possibly other places.

tuxoko:
I've already fixed that. It's a preexisting race in splat. It's just my
patch makes it almost always occur.

zio_wait is absolutely safe from this. Whether there exist such race, I'm
not 100 percent sure.

tuxoko:
@dweeezil 
Now I got to my desktop, I can explain more clearly.
The problem in splat code is that the waker (cv_broadcast, cv_signal) doing free (cv_destroy and mutex_destroy). This type of construct is always wrong, because you don't know when the waiter will release the mutex.

behlendorf:
@tuxoko @dweeezil the updated fix to include the test cases LGTM.  Either of you have reservations about merging this?

tuxoko:
@behlendorf 
I'd say go for it.

dweeezil:
@behlendorf No problems during high stress testing. LGTM.

behlendorf:
Great, thanks for quick reply.  Merged as:

e843553 Don't hold mutex until release cv in cv_wait

PR #520, opened 2016-01-08
_ILP32 is always defined on SPARC

F4S4K4N:
When using a SPARC CPU, _ILP32 will always be defined. This wont result in any issues when compiling 32bit, but when compiling 64bit you will end up with _ILP32 and _LP64 defined which will trigger the following error.

````C
#if defined(_ILP32) && defined(_LP64)
#error "Both _ILP32 and _LP64 are defined"
#endif
````

As you can see below, we already have a check for 32bit vs 64bit. So having _ILP32 always be defined not only creates issues for 64bit but is also needlessly redundant.

````C
#if defined(__arch64__)
#if !defined(_LP64)
#define _LP64
#endif
#else
#if !defined(_ILP32)
#define _ILP32
#endif
#endif
````

### DIFF
--- a/include/sys/isa_defs.h
+++ b/include/sys/isa_defs.h
@@ -118,10 +118,6 @@
 #define __sparc__
 #endif
 
-#if !defined(_ILP32)
-#define _ILP32
-#endif
-
 #if defined(__arch64__)
 #if !defined(_LP64)
 #define _LP64


F4S4K4N:
I forgot to mention, SPL compiles fine on SPARC with either 32bit or 64bit, but modules that compile against SPL, namely ZFS, will not unless you change the header as i have.

behlendorf:
@F4S4K4N could you open a pull request with this fix for SPARC.

On a related note this issue was able to slip in because we don't regularly test the build on SPARC/SPARC64.  Would happen to be in a position where you could contribute a SPARC/SPARC64 build slave?  All we need is to be able to run the buildbot slave as an unprivileged user on a system connected to the internet.  It can be located anywhere.  But this would make sure this doesn't happen again.

behlendorf:
Oh wait, this is a PR.  Thanks.  :)

F4S4K4N:
I have a build box in the works for a few other projects as well, i can give you guys access to it. It will probably be a little while until it's up though. Im on IRC almost all of the time under the same name.

behlendorf:
@F4S4K4N that sounds perfect.  If you can just drop me an email whenever it's ready I can set it up.  Or if you prefer I can provide directions and you can configure it.  Thanks.

PR #523, opened 2016-01-18
Turn on both PF_FSTRANS and PF_MEMALLOC_NOIO in spl_fstrans_mark

tuxoko:
In b4ad50a, we abandoned memalloc_noio_save in favor of spl_fstrans_mark
because earlier kernel with it doesn't turn off __GFP_FS. However, for newer
kernel, we would prefer PF_MEMALLOC_NOIO because it would work for allocation
in kernel which we cannot control otherwise. So in this patch, we turn on both
PF_FSTRANS and PF_MEMALLOC_NOIO in spl_fstrans_mark.

### DIFF
--- a/include/sys/kmem.h
+++ b/include/sys/kmem.h
@@ -78,14 +78,20 @@ typedef struct {
 	unsigned int saved_flags;
 } fstrans_cookie_t;
 
+#ifdef PF_MEMALLOC_NOIO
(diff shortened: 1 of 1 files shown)


behlendorf:
LGTM.  I was originally concerned that not all PF_FSTRANS consumers, like those in spl-vnode.c`, were changed but upon review this looks right.  Those check should solely concern themselves with PF_FSTRANS.

PR #525, opened 2016-01-20
Create spl-kmod-debuginfo rpm with redhat spec file

ofaaland:
Correct the redhat specfile so that working debuginfo rpms are created
for the kernel modules.  The generic specfile already does the right
thing.

SPL equivalent of zfsonlinux/zfs Issue 4224


behlendorf:
Merged as:

7323da1 Create spl-kmod-debuginfo rpm with redhat spec file

PR #527, opened 2016-01-26
make sed more discriminating to ensure spl/ only occurs once in core-y

### LINKED ISSUE #526: copy-builtin cannot be run more than once against the same target tree
line 119 of copy-builtin:
```sed -i 's#+= kernel/#+= kernel/ spl/#' "$KERNEL_DIR/Makefile"```
results in line 887 of /usr/src/linux/Makefile:
```core-y      += kernel/ spl/ spl/ mm/ fs/ ipc/ security/ crypto/ block/```
upon make:
```Makefile:946: target 'spl' given more than once in the same rule````

something like
```grep 'kernel/ spl/' "$KERNEL_DIR/Makefile" || sed -i 's#+= kernel/#+= kernel/ spl/#' "$KERNEL_DIR/Makefile"```
is completely inelegant, but would fix this.

infowolfe:
This fixes #526

### DIFF
--- a/copy-builtin
+++ b/copy-builtin
@@ -116,7 +116,7 @@ add_after()
 add_after "$KERNEL_DIR/Kconfig" 'source "arch/$SRCARCH/Kconfig"' 'source "spl/Kconfig"'
 # We must take care to build SPL before ZFS, otherwise the symbols required
 # to link ZFS will not be available.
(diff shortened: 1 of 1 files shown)


behlendorf:
Merged as:

d112232 Ensure spl/ only occurs once in core-y

PR #528, opened 2016-01-28
Do irqsave in tsd to prevent lockup

tuxoko:
taskq_member could be called from interrupt context via zio_execute. We should
do irqsave in tsd to prevent lockup.


tuxoko:
I think I misunderstood the comment in zio_execute. It seems interrupt context will dispatch a task to go into zio_execute, so this probably isn't needed.

behlendorf:
I was going to ask if you had a back trace or something showing this problem.  Before we go around disabling interrupts anywhere unnecessarily let's make sure we have a smoking gun so to speak.

PR #529, opened 2016-01-28
Allow kicking a taskq to spawn more threads

tuxoko:
This patch add a module parameter spl_taskq_kick. When writing non-zero value
to it, it will scan all the taskq, if a taskq contains a task pending for more
than 5 seconds, it will be forced to spawn a new thread. This is use as an
emergency recovery from deadlock, not a general solution.

### DIFF
--- a/include/sys/taskq.h
+++ b/include/sys/taskq.h
@@ -105,6 +105,7 @@ typedef struct taskq_ent {
 	void			*tqent_arg;
 	taskq_t			*tqent_taskq;
 	uintptr_t		tqent_flags;
+	unsigned long		tqent_birth;
 } taskq_ent_t;
 
 #define	TQENT_FLAG_PREALLOC	0x1
--- a/man/man5/spl-module-parameters.5
+++ b/man/man5/spl-module-parameters.5
@@ -253,6 +253,21 @@ Default value: \fB/etc/hostid\fR
 .sp
 .ne 2
 .na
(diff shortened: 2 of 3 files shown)


tuxoko:
Partial fix for https://github.com/zfsonlinux/zfs/issues/4169

behlendorf:
This looks like a reasonable way to partially address this unlikely issue for now.  You'll want to add an entry to the spl-module-parameter(5) for this new option.

tuxoko:
Wait for https://github.com/zfsonlinux/zfs/pull/4283 to build test.

tuxoko:
Finally it built. I gave up on using a compat layer of module_param_cb on old kernel. I don't think it's possible.

tuxoko:
@behlendorf 
So for the current linux master, `module_param_call` and `module_param_cb` has about the same number of users. So I think `module_param_call` being "obsolete" for all the years most likely won't disappear soon. If we are going to use this interface in other places. I think it's better to use `module_param_call`. Because it's much easier to have a _call compat layer than _cb compat layer.

behlendorf:
That sounds reasonable to me.  It would be best if we could lay that infrastructure now, but if in practice the whole thing ends up too convoluted I don't have an objection to how you've handled it in this patch.

behlendorf:
Merged with minor update to the man page.

8f3b403 Allow kicking a taskq to spawn more threads

PR #530, opened 2016-02-08
Allow spawning a new thread for TQ_NOQUEUE dispatch with dynamic taskq

dweeezil:
When a TQ_NOQUEUE dispatch is done on a dynamic taskq, allow another
thread to be spawned.  This will cause TQ_NOQUEUE to behave similarly
as it does with non-dynamic taskqs.

### DIFF
--- a/module/spl/spl-taskq.c
+++ b/module/spl/spl-taskq.c
@@ -562,16 +562,22 @@ taskq_dispatch(taskq_t *tq, task_func_t func, void *arg, uint_t flags)
 
 	/* Do not queue the task unless there is idle thread for it */
 	ASSERT(tq->tq_nactive <= tq->tq_nthreads);
-	if ((flags & TQ_NOQUEUE) && (tq->tq_nactive == tq->tq_nthreads))
-		goto out;
+	if ((flags & TQ_NOQUEUE) && (tq->tq_nactive == tq->tq_nthreads)) {
+		/* Dynamic taskq may be able to spawn another thread */
+		if (!(tq->tq_flags & TASKQ_DYNAMIC) || taskq_thread_spawn(tq) == 0)
+			goto out;
+	}
 
 	if ((t = task_alloc(tq, flags, &irqflags)) == NULL)
 		goto out;
 
 	spin_lock(&t->tqent_lock);
 
+	/* Queue to the front of the list to enforce TQ_NOQUEUE semantics */
+	if (flags & TQ_NOQUEUE)
+		list_add(&t->tqent_list, &tq->tq_prio_list);
 	/* Queue to the priority list instead of the pending list */
-	if (flags & TQ_FRONT)
+	else if (flags & TQ_FRONT)
 		list_add_tail(&t->tqent_list, &tq->tq_prio_list);
 	else
 		list_add_tail(&t->tqent_list, &tq->tq_pend_list);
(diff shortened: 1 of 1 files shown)


dweeezil:
Passes splat taskq tests but otherwise untested at the moment.

tuxoko:
@dweeezil 
I'm wondering if the TQ_NOQUEUE semantic needs to be strict, because what we are currently doing is not strict at all. Even if `tq->tq_nactive < tq->tq_nthreads`, there might be pending task on the list. So in the end it will still be queued.

tuxoko:
@dweeezil 
You also need to prevent doing 2 spawns.

behlendorf:
Here's the contract as described by the Solaris documentation.  My interpretation of this is we can correctly return NULL for this for if we need too and the caller must handle it.  If we can avoid doing so such much the better, but it's a legitimate things the caller needs to handle.

```
TQ_NOQUEUE: Do not enqueue a task if it can't be dispatched due to lack
of available resources and return NULL. If this flag is not set and the task
pool is exhausted, the task may be scheduled in backing task queue. This
flag should always be used when a task queue is used for tasks that may
depend on each other for completion. Enqueueing dependent tasks may
create deadlocks. 

TQ_SLEEP: Do not wait for resources; may return NULL. 

TQ_NOSLEEP: May block waiting for resources. May still fail for dynamic
task queues if TQ_NOQUEUE is also specified.
```

dweeezil:
@tuxoko Good point, likely no need to spawn twice.

@behlendorf That's more or less what I expected to see in the documentation (which I had never bothered actually looking up).

After thinking about this a bit more, it seems we probably need to add ```TQ_NOQUEUE``` tqents to the _head_ of ```tq_pend_list```.  Otherwise if there's already something pending, then it (the existing pend) may get the thread rather than the one being requested with ```TQ_NOQUEUE```.  Furthermore, it seems the attempts to allocate the tqent ought to come before the spawn.

And I suppose even adding the tqent to the head of the pending list isn't good enough because there still may be a race, albeit unlikley, where another ```TQ_NOQUEUE``` entry could be dispatched and grab the thread which was already "reserved".  It almost seems as if ```TQ_NOQUEUE``` is almost like a "super priority" mode and might deserve its own pending list.

Finally, I'll note that although I've been aware that the SPL has not had the correct behavior in this case since dynamic taskqs were introduced, it has come to a head due to the new use of ```TQ_NOQUEUE``` in zfsonlinux/zfs#3656.  I'm still hoping to determine whether it  _really_ needs to use ```TQ_NOQUEUE``` in the first place.

dweeezil:
I'm not sure why I thought the tqent allocation should come before the initial spawn.

dweeezil:
It looks like splat's taskq test 6 can be enhanced pretty easily to test ```TQ_NOQUEUE``` with dynamic taskqs.

behlendorf:
@dweeezil one possible way to close the remaining race you mentioned (which is very unlikely) is to keep a count of how many TQ_NOQUEUE tasks are the head of the pending queue.  That would be your reserved count to add in when determine if it could use the threads.  However, I'm not sure it's worth the added complexity.

You'll want to add this same logic to `taskq_dispatch_ent()`, and I'm all for extending the taskq test cases.

dweeezil:
@behlendorf I was planning on dealing with ```taskq_dispatch_ent()``` and also extending the taskq splat test coverage.

dweeezil:
Added ```TQ_NOQUEUE``` support to ```taskq_dispatch_ent()```.  I'm thinking maybe we ought to have an ASSERT in ```taskq_dispatch_delay()``` to prohibit ```TQ_NOQUEUE``` for the time being.  It would be difficult to implement it in a sensible way without a concept of "reserved" threads (which would also be a way of working around the small remaining race discussed previously).

behlendorf:
LGTM and I think adding the ASSERT to `taskq_dispatch_delay()` is a good idea.

behlendorf:
@dweeezil sorry about the delay.  Merged as:

7bb5d92 Allow spawning a new thread for TQ_NOQUEUE dispatch with dynamic taskq

PR #532, opened 2016-02-12
Use raw_smp_processor_id instead of smp_processor_id

ryao:
This means callers do not need to turn off preemption in places where
using this information is safe. Consequently, there is no need to modify
consumers to call `kpreempt_disable()` and `kpreempt_enable()`, which
means that we do not insert a `preempt_schedule()` into places where
there is none on Illumos.

### DIFF
--- a/include/sys/processor.h
+++ b/include/sys/processor.h
@@ -25,7 +25,7 @@
 #ifndef	_SPL_PROCESSOR_H
 #define	_SPL_PROCESSOR_H
 
-#define getcpuid() smp_processor_id()
+#define getcpuid() raw_smp_processor_id()
 
 typedef int	processorid_t;
 
--- a/include/sys/sysmacros.h
+++ b/include/sys/sysmacros.h
@@ -79,7 +79,7 @@
 #define curproc				current
 #define max_ncpus			num_possible_cpus()
 #define boot_ncpus			num_online_cpus()
-#define CPU_SEQID			smp_processor_id()
+#define CPU_SEQID			raw_smp_processor_id()
 #define _NOTE(x)
 #define is_system_labeled()		0
 


tuxoko:
@ryao I don't understand your commit message.
And why do we want this? `smp_processor_id` only differs from `raw_smp_processor_id` when you do CONFIG_DEBUG_PREEMPT

behlendorf:
@tuxoko I believe this is just so we could remove some `kpreempt_disable()` and `kpreempt_enable()` wrappers in the ZFS code.  There are a few places in the existing code where `getcpuid()` is used as a lazy way to just get a small random value for indexing.

How about adding our own raw variant rather than disabling the (absolutely correct) preemption warnings in `getcpuid()`.  This way you're still able to drop the `kpreempt_disable()` and `kpreempt_enable()` wrappers in the ZFS code and it's clear your getting a raw value which may change.

ryao:
@tuxoko The purpose of this is to silence warnings when `getcpuid()` or `CPU_SEQID` is used in a manner that might be incorrect. Such warnings are spurious because our usage is always correct.

@behlendorf The `getcpuid()` usage is in those places is more than a simple index. It is an attempt to select a resource out of several resources that has the lowest probability of being contended. Unless preemption of something else holding it occurred, the next pass through the code using this technique would grab the resource with no locking issues and proceed with minimal latency. Calling `kpreempt_enable()` after doing that introduces a voluntary preemption point on both Linux and illumos-gate, which increases the likelihood of contention because it means that we have grabbed the resource and are now explicitly asking the scheduler to consider scheduling another thread that potentially wants it too.

I do not think adding our own raw variant makes sense because the illumos versions are semantically equivalent to `raw_smp_processor_id()`. Anyone familiar with illumos development would know the semantics of these are raw while anyone who does not would know from checking the definitions in the SPL. Also, identifying incorrect usage at runtime is a task for which the buildbot should be well suited.

behlendorf:
losing.  I'd prefer not to merge this until we have a more compelling reason to do so.

Issue #534, opened 2016-03-06
need a working rw_tryupgrade()


dweeezil:
The lack of a working ```rw_tryupgrade()``` in all but the ```CONFIG_RWSEM_GENERIC_SPINLOCK``` case (which is pretty rare these days) can have ill effects if the upstream OpenZFS code uses it in certain performance-sensitive areas.

For example, zfsonlinux/zfs@7f60329 introduced a statement which inhibits creation of new prefetch streams when ```rw_tryupgrade()``` returns zero.

The problem stems from the imperfect mapping between the Solaris ```krwlock``` facility and Linux's ```rw_semaphore``` facility (see d28db80 and others for details).  In addition to the differing API, Linux does not natively support an upgrade from a reader lock to a reader/writer lock (but it does support the reverse).

Since the entire Solaris krwlock facility is implemented as macros which expand to statement expressions and also since the krwlock_t structure is a wrapper for rw_semaphore, there's a great deal of flexibility available within the SPL.  It seems that, for example, it might be possible to add additional members to ```krwlock_t``` to help track such things as pending waiters for write or any other accounting and/or spinlocks which might be useful to implement a semantically equivalent ```rw_tryupgrade()```.

[EDIT] Referencing zfsonlinux/zfs#4388.

dweeezil:
After auditing the other users of ```rw_tryupgrade()``` in ZFS, it doesn't seem this is worth doing at the moment.  The code mentioned in zfsonlinux/zfs@7f60329 is the only place in which it's used without effectively immediately dropping and re-acquiring a RW lock in the failure case.

dweeezil:
Closing.  As indicated in https://github.com/zfsonlinux/zfs/issues/4388#issuecomment-193538479 and also from my own audit, this isn't worth implementing, but at least it's mentioned here for documentation in the future.

PR #536, opened 2016-03-09
Add rw_tryupgrade()

behlendorf:
This implementation of rw_tryupgrade() behaves slightly differently
from its counterparts on other platforms.  It drops the RW_READER lock
and then acquires the RW_WRITER lock leaving a small window where no
lock is held.  On other platforms the lock is never released during
the upgrade process.  This is necessary under Linux because the kernel
does not provide an upgrade function.

There are currently no callers in the ZFS code where this change in
behavior is a problem.  In fact, in most cases the code is already
written such that if the upgrade fails the RW_READER lock is dropped
and the caller blocks waiting to acquire the lock as RW_WRITER.

Issue zfsonlinux/zfs#4388
Issue #534


prometheanfire:
What additional testing is needed for this?

dweeezil:
LGTM, pretty much as expected.  The pr does seem to have survived the 'bots.

behlendorf:
@prometheanfire if you could independently verify it resolves the prefetch issue that would be great.

@dweeezil thanks for the quick review.

prometheanfire:
tested via scrub on a 10G mirror (over iscsi attached volumes using nova/cinder)

without: 276M/s (completed in less than a minute)
with patch: 279M/s (less than a minute)

guess I'll try with fio?

dweeezil:
@prometheanfire You want to look at ```/proc/spl/kstat/zfs/zfetchstats```.

prometheanfire:
with

    name                            type data
    hits                            4    238
    misses                          4    64112
    max_streams                     4    63163

without

    name                            type data
    hits                            4    0
    misses                          4    50996
    max_streams                     4    0

behlendorf:
Looks good.  OK, then I'll get this merged and add @dweeezil's signed-off-by.  @prometheanfire thanks for the confirmation.

behlendorf:
Merged as:

a6ae97c Add rw_tryupgrade()

PR #537, opened 2016-03-16
Add support for s390[x].

xnox:
Signed-off-by: Dimitri John Ledkov <xnox@ubuntu.com>

### DIFF
--- a/include/sys/isa_defs.h
+++ b/include/sys/isa_defs.h
@@ -131,7 +131,21 @@
 #define _BIG_ENDIAN
 #define _SUNOS_VTOC_16
 
-#else /* Currently x86_64, i386, arm, powerpc, and sparc are supported */
+/* s390 arch specific defines */
+#elif defined(__s390__)
+#if defined(__s390x__)
+#if !defined(_LP64)
+#define	_LP64
+#endif
+#else
+#if !defined(_ILP32)
+#define	_ILP32
+#endif
+#endif
+
+#define	_BIG_ENDIAN
+
+#else /* Currently x86_64, i386, arm, powerpc, s390, and sparc are supported */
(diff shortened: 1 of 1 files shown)


xnox:
I have tested this on s390x, but not on s390 as i don't have such a system/config/compiler. But it should be correct, w.r.t. defines.

behlendorf:
@xnox this looks reasonable to me, although I wish I had a way to verify it.  When you open the ZFS counterpart for this patch which modifies `lib/libspl/include/sys/isa_defs.h` similarly please add the following line to reference this PR.  That will cause the buildbot to automatically test both of these patches.

```
Requires-spl: refs/pull/537/head
```

ryao:
@behlendorf Verifying it can be done by installing a Linux distribution that supports the z/architecture inside the Hercules emulator and then ZoL inside of that.

xnox:
actually i believe VTOC should not be defined here, but is appropriate for the zfs repo.

ryao:
@xnox We will need either _SUNOS_VTOC_16 or _SUNOS_VTOC_8 defined. Otherwise, `lib/libspl/include/sys/dklabel.h` will trigger a build failure from `#error No VTOC format defined.`. Everyone else defined `_SUNOS_VTOC_16`, so we probably could get away with defining that too.

xnox:
@ryao yes, which is what i did in the zfs copy of libspl. There are two copies of this code as far as i can see. And in zfs repo more arches define  _SUNOS_VTOC_16, unlike in this repo.

ryao:
@xnox You are correct. I mistook this for the issue in the zfs repository.

behlendorf:
Merged as:

224817e Add support for s390[x].

PR #540, opened 2016-04-12
Use kernel provided mutex owner

tuxoko:
To reduce mutex footprint, we detect the existence of owner in kernel mutex,
and rely on it if it exists.

Note that before Linux 3.0, mutex owner is of type thread_info. Also note
that, in Linux 3.18, the condition for owner is changed from
CONFIG_DEBUG_MUTEXES || CONFIG_SMP to
CONFIG_DEBUG_MUTEXES || CONFIG_MUTEX_SPIN_ON_OWNER

### DIFF
--- a/config/spl-build.m4
+++ b/config/spl-build.m4
@@ -44,6 +44,7 @@ AC_DEFUN([SPL_AC_CONFIG_KERNEL], [
 	SPL_AC_USLEEP_RANGE
 	SPL_AC_KMEM_CACHE_ALLOCFLAGS
 	SPL_AC_WAIT_ON_BIT
+	SPL_AC_MUTEX_OWNER
 ])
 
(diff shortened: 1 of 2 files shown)


behlendorf:
So not that long ago we retired similar functionality in commit 5f920fbee1fd1838bad2d5a262041a3a171b6834 due to unwanted complexity.  That said, looking at your patch it certainly looks like we've refactored the code enough that that's no longer a major concern.  Particularly if we don't try to handle older kernels.

tuxoko:
Right..., I don't even remember we use kernel owner before. And I added the same thing for rwlock, so I was wondering why we don't do this already.
While reducing one pointer doesn't seems much, there are some structures like znode have multiple mutexes embedded. And if you consider how much znodes a host could have.

behlendorf:
Merged as:

cdd39dd Use kernel provided mutex owner

PR #542, opened 2016-04-18
Disable PF_FSTRANS flag when calling into fallocate

lorddoskias:
When spl calls into the backing's filesystem fallocate function it is doing
this with PF_FSTRANS set. This is fine in cases when fallocate triggers writeback
and the backing file system is XFS. In such cases a warning is triggered in XFS.

To fix it, just disable PF_FSTRANS flag in vn_space before calling into the
backing filesystem

Signed-off-by: Nikolay Borisov <kernel@kyup.com>


behlendorf:
Closing is favor of #543.  @lorddoskias if you could review the version of your fix in #543 we can work on getting it merged.

PR #543, opened 2016-04-25
Use vmem_free() in dfl_free() and add dfl_alloc()

dweeezil:
This change was lost, somehow, in e5f9a9a.  Since the arrays can be
rather large, they need to be allocated with vmem_zalloc() and freed
with vmem_free().

### DIFF
--- a/include/sys/dkioc_free_util.h
+++ b/include/sys/dkioc_free_util.h
@@ -48,7 +48,11 @@ typedef struct dkioc_free_list_s {
 } dkioc_free_list_t;
 
 static inline void dfl_free(dkioc_free_list_t *dfl) {
-	kmem_free(dfl, DFL_SZ(dfl->dfl_num_exts));
+	vmem_free(dfl, DFL_SZ(dfl->dfl_num_exts));
+}
+
+static inline dkioc_free_list_t *dfl_alloc(uint64_t dfl_num_exts, int flags) {
+	return vmem_zalloc(DFL_SZ(dfl_num_exts), flags);
 }
 
 #endif /* _SPL_DKIOC_UTIL_H */
--- a/module/spl/spl-vnode.c
+++ b/module/spl/spl-vnode.c
@@ -571,6 +571,9 @@ int vn_space(vnode_t *vp, int cmd, struct flock *bfp, int flag,
     offset_t offset, void *x6, void *x7)
 {
 	int error = EOPNOTSUPP;
+#ifdef FALLOC_FL_PUNCH_HOLE
+	int fstrans;
+#endif
 
 	if (cmd != F_FREESP || bfp->l_whence != 0)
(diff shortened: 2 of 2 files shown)


behlendorf:
@dweeezil I must have accidentally applied a stale version e5f9a9a when this was merged.  LGTM.

As for 9ac4097 see my question in zfsonlinux/zfs#4529.

include/sys/dkioc_free_util.h
```diff
@@ -48,7 +48,7 @@ typedef struct dkioc_free_list_s {
 } dkioc_free_list_t;
 
 static inline void dfl_free(dkioc_free_list_t *dfl) {
-	kmem_free(dfl, DFL_SZ(dfl->dfl_num_exts));
+	vmem_free(dfl, DFL_SZ(dfl->dfl_num_exts));
```
behlendorf: Shouldn't there be a `dfl_alloc()` counterpart to prevent this kind of issue.  I see it's directly `vmem_alloc()` in the trim code.

dweeezil: It would make perfect sense.  I think the only reason we don't have one is because illumos doesn't have one.  I do like the idea of adding one.  We've already got the kmem/vmem difference from the upstream and I suppose a ```dfl_alloc()``` would make the intention more obvious for future porting efforts.

behlendorf: Then let's add one.  We're already forced to diverge slightly here, let's diverge in a way which can in theory be pushed back upstream.

behlendorf:
LGTM.  My only concern would be if we need to leave PF_FSTRANS clear over `truncate_range()`.  However, it looks like it's safe to leave it set.

lorddoskias:
The "disable PF_FSTRANS" portions LGTM. However, I'd like to make it clear that the issue stems from the portion of XFS which is writing pages to disk, in my bug report it just so happened that page write code was invoked from fallocate. Having patched the fallocate callsite doesn't mean such warnings won't occur in the future. In any case, I'm happy with this version being merged.

dweeezil:
Regarding the ```truncate_range()``` callback, according to torvalds/linux@17cf28afea2a1112f240a3a2da8af883be024811, it was only ever supported by tmpfs.  It didn't appear to me that tmpfs ever cared about the state of ```PF_FSTRANS```.

As to the commit comment, I mainly cloned the one from 2a5d574.  I suppose some different wording could be in order.

behlendorf:
OK, then I'll merge them both as is.  I can add additional context to the commit message to make it clear under exactly what context this can occur and link back to the issue.

behlendorf:
Merged as:

ea2633a Clear PF_FSTRANS over spl_filp_fallocate()
3bf657b Use vmem_free() in dfl_free() and add dfl_alloc()

PR #548, opened 2016-05-11
Add cv_timedwait_sig_hires to allow interruptible sleep

tuxoko:

### DIFF
--- a/module/spl/spl-condvar.c
+++ b/module/spl/spl-condvar.c
@@ -26,6 +26,7 @@
 
 #include <sys/condvar.h>
 #include <sys/time.h>
+#include <linux/hrtimer.h>
 
 void
 __cv_init(kcondvar_t *cvp, char *name, kcv_type_t type, void *arg)
@@ -238,7 +239,7 @@ __cv_timedwait_hires(kcondvar_t *cvp, kmutex_t *mp, hrtime_t expire_time,
 	DEFINE_WAIT(wait);
 	kmutex_t *m;
 	hrtime_t time_left, now;
-	unsigned long time_left_us;
+	ktime_t ktime_left;
 
 	ASSERT(cvp);
 	ASSERT(mp);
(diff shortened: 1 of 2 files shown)


module/spl/spl-condvar.c
```diff
@@ -312,10 +312,27 @@ cv_timedwait_hires(kcondvar_t *cvp, kmutex_t *mp, hrtime_t tim, hrtime_t res,
...
+cv_timedwait_hires(kcondvar_t *cvp, kmutex_t *mp, hrtime_t tim, hrtime_t res,
+    int flag)
+{
+	return (cv_timedwait_hires_common(cvp, mp, tim, res, flag,
+	    TASK_UNINTERRUPTIBLE));
 }
 EXPORT_SYMBOL(cv_timedwait_hires);
 
+hrtime_t
+cv_timedwait_hires_sig(kcondvar_t *cvp, kmutex_t *mp, hrtime_t tim, hrtime_t res,
```
behlendorf: We should call this `cv_timedwait_sig_hires()` which is consistent with what it's called in [illumos](https://github.com/openzfs/openzfs/blob/master/usr/src/uts/common/os/condvar.c#L363).

tuxoko: done.

module/spl/spl-condvar.c
```diff
@@ -231,7 +231,7 @@ EXPORT_SYMBOL(__cv_timedwait_sig);
  * 'expire_time' argument is an absolute clock time in nanoseconds.
  * Return value is time left (expire_time - now) or -1 if timeout occurred.
  */
-static clock_t
+static hrtime_t
 __cv_timedwait_hires(kcondvar_t *cvp, kmutex_t *mp, hrtime_t expire_time,
```
behlendorf: This bugs me too but the illumos version of this does return a `clock_t`.  However, it looks like have a longer standing bug where we're not converting the hrtime_t to a clock_t with `NSEC_TO_TICK().`

tuxoko: We don't seems to use return value anyway.

behlendorf: So at the moment it's harmless but let's preserve the existing upstream prototype, and even better for the return value.

tuxoko: Right.

behlendorf:
LGTM

tuxoko:
Change usleep_range to schedule_hrtimeout_range

PR #549, opened 2016-05-18
Linux 4.7 compat: inode_lock() and friends

tuxoko:
Linux 4.7 changes i_mutex to i_rwsem, and we should used inode_lock and
inode_lock_shared to do exclusive and shared lock respectively.

We use spl_inode_lock{,_shared}() to hide the difference. Note that on older
kernel you'll always take an exclusive lock.

### DIFF
--- a/config/spl-build.m4
+++ b/config/spl-build.m4
@@ -45,6 +45,7 @@ AC_DEFUN([SPL_AC_CONFIG_KERNEL], [
 	SPL_AC_KMEM_CACHE_ALLOCFLAGS
 	SPL_AC_WAIT_ON_BIT
 	SPL_AC_MUTEX_OWNER
+	SPL_AC_INODE_LOCK
 ])
 
 AC_DEFUN([SPL_AC_MODULE_SYMVERS], [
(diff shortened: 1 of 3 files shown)


include/linux/file_compat.h
```diff
@@ -76,8 +76,17 @@ spl_filp_fallocate(struct file *fp, int mode, loff_t offset, loff_t len)
...
 
+#ifdef HAVE_INODE_LOCK_SHARED
+#define	spl_inode_lock(ip)		inode_lock_nested(ip, I_MUTEX_PARENT)
+#define	spl_inode_unlock(ip)		inode_unlock(ip)
+#define	spl_inode_lock_shared(ip)	inode_lock_shared(ip)
+#define	spl_inode_unlock_shared(ip)	inode_unlock_shared(ip)
+#else
 #define	spl_inode_lock(ip)		mutex_lock_nested(&(ip)->i_mutex, I_MUTEX_PARENT)
 #define	spl_inode_unlock(ip)		mutex_unlock(&(ip)->i_mutex)
+#define	spl_inode_lock_shared(ip)	mutex_lock(&(ip)->i_mutex)
```
behlendorf: The shared version should also use `mutex_lock_nested()` and pass I_MUTEX_PARENT for the reason listed in 628fc52137fc14377eba46c66b57d8d094e88507.  This way we won't get lockdep warnings on older kernels running newer versions of ZFS which use `spl_inode_lock_shared()`.

config/spl-build.m4
```diff
@@ -1476,3 +1477,26 @@ AC_DEFUN([SPL_AC_MUTEX_OWNER], [
 	])
 	EXTRA_KCFLAGS="$tmp_flags"
 ])
+
+dnl #
+dnl # 4.7 API change
+dnl # i_mutex is changed to i_rwsem. Instead of directly using
+dnl # i_mutex/i_rwsem, we should use inode_lock() and inode_lock_shared()
```
behlendorf: Technically `inode_lock()` was introduced in 4.5 and `inode_lock_shared()` in 4.7.  But I think it does make sense to simply this and group them since if they have `inode_lock_shared()` they'll also have `inode_lock()`.

include/linux/file_compat.h
tuxoko: We currently have only one user of spl_inode_lock_shared(llseek hole), and it's not nested so it should fine. Hard coding I_MUTEX_PARENT looks like a hacky work around anyway, so we should leave it alone until we need it.

behlendorf: For the moment it's fine, but I suspect it will bite us some day.  This raises a related point, it would be prudent to establish `spl_` wrappers now for all of the inode lock variants to avoid future potential spl/zfs compatibility issues.

```
spl_inode_lock
spl_inode_unlock
spl_inode_lock_shared
spl_inode_unlock_shared
spl_inode_trylock
spl_inode_trylock_shared
spl_inode_is_locked
spl_inode_lock_nested
```

tuxoko: It's an interesting thing that there's no `inode_lock_shared_nested`. Perhaps, they haven't found out a need to nest a shared lock. So for consistency, I'd still want to keep `spl_inode_lock_shared` non-nested.

Do you think I should map `spl_inode_lock` to non-nested and change the required users to spl_inode_lock_nested?

behlendorf: Yeah, I think now would be the right time to make that change.

tuxoko:
Let's hope they don't bother to change I_MUTEX_* to I_RWSEM_*

behlendorf:
Agreed.  This LGTM.

PR #551, opened 2016-05-20
Improve spl slab cache alloc

jxiong:
The policy is to try to allocate with KM_NOSLEEP, which will lead to
memory allocation with GFP_ATOMIC, and if it fails, it will launch
an taskq to expand slab space.

This way it should be able to get better NUMA memory locality and
reduce the overhead of context switch.

Signed-off-by: Jinshan Xiong <jinshan.xiong@gmail.com>

### DIFF
--- a/module/spl/spl-kmem-cache.c
+++ b/module/spl/spl-kmem-cache.c
@@ -1149,15 +1149,13 @@ spl_cache_obj(spl_kmem_cache_t *skc, spl_kmem_slab_t *sks)
  * It is responsible for allocating a new slab, linking it in to the list
  * of partial slabs, and then waking any waiters.
  */
-static void
-spl_cache_grow_work(void *data)
+static int
+__spl_cache_grow(spl_kmem_cache_t *skc, int flags)
 {
-	spl_kmem_alloc_t *ska = (spl_kmem_alloc_t *)data;
-	spl_kmem_cache_t *skc = ska->ska_cache;
 	spl_kmem_slab_t *sks;
 
 	fstrans_cookie_t cookie = spl_fstrans_mark();
-	sks = spl_slab_alloc(skc, ska->ska_flags);
+	sks = spl_slab_alloc(skc, flags);
 	spl_fstrans_unmark(cookie);
 
 	spin_lock(&skc->skc_lock);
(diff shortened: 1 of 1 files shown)


behlendorf:
So for lack of a better term it sounds like you're trying to add a fast path for new slab allocation.  If I understand correctly this is designed to improve the case where the slab being allocated is small or the case where there's lots of free memory.  Since the allocation is KM_NOSLEEP it's expected to succeed or fail quickly then fall-back to allocating in the taskq.  That sounds reasonable.

My only concern is that the way the code is written, calling `__spl_cache_grow()` might end up in a blocking vmalloc allocation.  Specifically `__spl_cache_grow()->spl_slab_alloc()->kv_alloc()->vmalloc()`.  This can result in a deadlock in `vmalloc()` which is why it's always done in a taskq thread.

The fast path idea is good but only for KMC_KMEM backed slab caches.  It probably makes the most sense to also update `kv_alloc()` to fail immediately with NULL when KM_NOSLEEP is set for !KMC_KMEM backed caches.

When you make this change also make sure to rebase this patch against the latest master SPL source and we'll be able to run it through the automated testing.

module/spl/spl-kmem-cache.c
```diff
@@ -1213,6 +1225,10 @@ spl_cache_grow(spl_kmem_cache_t *skc, int flags, void **obj)
 		return (rc ? rc : -EAGAIN);
 	}
 
+	rc = __spl_cache_grow(skc, flags | KM_NOSLEEP);
+	if (rc == 0)
+		return (rc);
```
behlendorf: It's functionally identical but I think a `return (0)` here would flow better.  Also please add a comment describing the motivation behind adding this KM_NOSLEEP allocation.

jxiong:
> Specifically __spl_cache_grow()->spl_slab_alloc()->kv_alloc()->vmalloc(). This can result in a deadlock in vmalloc() which is why it's always done in a taskq thread.

In that case, we can use __vmalloc() and pass in GFP_ATOMIC.

It could be done better by returning the objects to the NUMA node where it was allocated, which I have done in the ABD patches. But even with this simple patch, it's already much better than before.

behlendorf:
@jxiong unfortunately `__vmalloc()` doesn't honor GFP_ATOMIC.  It relies on `map_vm_area()` which doesn't use the passed flags.  That means it'll appear to work correctly virtually all of the time but once in the while you'll get unlucky.  See https://bugzilla.kernel.org/show_bug.cgi?id=30702

It would be safe to use the patch with kmem backed slabs and we still might be most of the benefit.

behlendorf:
@jxiong thanks for refreshing this.  The updated patch LGTM and I've opened a [dummy PR](https://github.com/zfsonlinux/zfs/pull/4678) on the ZFS tracker to run this patch through the buildbot.

behlendorf:
Merged as:

16fc1ec Improve spl slab cache alloc

PR #553, opened 2016-05-21
SPL fixes

tuxoko:
1. Restore CALLOUT_FLAG_ABSOLUTE in cv_timedwait_hires
2. Fix splat-condvar on Linux 4.7
3. Fix race in taskq_destroy and dynamic spawing

### DIFF
--- a/module/spl/spl-condvar.c
+++ b/module/spl/spl-condvar.c
@@ -311,9 +311,8 @@ cv_timedwait_hires_common(kcondvar_t *cvp, kmutex_t *mp, hrtime_t tim, hrtime_t 
 		tim = (tim / res) * res;
 	}
 
-	ASSERT(!(flag & CALLOUT_FLAG_ABSOLUTE));
-	/* get abs expire time */
-	tim += gethrtime();
+	if (!(flag & CALLOUT_FLAG_ABSOLUTE))
+		tim += gethrtime();
 
 	return (__cv_timedwait_hires(cvp, mp, tim, state));
 }
--- a/module/spl/spl-taskq.c
+++ b/module/spl/spl-taskq.c
@@ -447,8 +447,8 @@ taskq_wait_outstanding_check(taskq_t *tq, taskqid_t id)
 void
 taskq_wait_outstanding(taskq_t *tq, taskqid_t id)
 {
-	wait_event(tq->tq_wait_waitq,
-	    taskq_wait_outstanding_check(tq, id ? id : tq->tq_next_id - 1));
+	id = id ? id : tq->tq_next_id - 1;
+	wait_event(tq->tq_wait_waitq, taskq_wait_outstanding_check(tq, id));
 }
 EXPORT_SYMBOL(taskq_wait_outstanding);
 
--- a/module/spl/spl-taskq.c
+++ b/module/spl/spl-taskq.c
@@ -763,11 +763,12 @@ taskq_thread_spawn_task(void *arg)
 	taskq_t *tq = (taskq_t *)arg;
 	unsigned long flags;
 
-	(void) taskq_thread_create(tq);
-
-	spin_lock_irqsave_nested(&tq->tq_lock, flags, tq->tq_lock_class);
-	tq->tq_nspawn--;
-	spin_unlock_irqrestore(&tq->tq_lock, flags);
+	if (taskq_thread_create(tq) == NULL) {
+		/* restore spawning count if failed */
+		spin_lock_irqsave_nested(&tq->tq_lock, flags, tq->tq_lock_class);
+		tq->tq_nspawn--;
+		spin_unlock_irqrestore(&tq->tq_lock, flags);
+	}
 }
 
 /*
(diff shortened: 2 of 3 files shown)


dweeezil:
Looking at the taskq_destroy/spawning issue only:  I'm wondering whether you really need tqt_is_dynamic.  It would seem that taskq_destroy could only do the wait-for-nspawn==0 loop if TASKQ_DYNAMIC were set.  Similarly the taskq_thread could only decrement nspawn if TASKQ_DYNAMIC is set although it ought to be harmless to decrement it no matter what since nspawn should only be looked at for a TASKQ_DYNAMIC taskq.

tuxoko:
Well, taskq_create will create one thread even if it's TASKQ_DYNAMIC.

dweeezil:
@tuxoko Interesting you should mention the single thread created for dynamic taskqs.  There doesn't seem to be much of a need them to have a thread at all.  What are your thoughts on 345f9be which changes things a few things in your patch?  I've run the splat tests, ztest and a few hours (so far) of the test suite with it.

tuxoko:
@dweeezil 
Spawning one thread is probably and old relics before we spawn thread during taskq_dispatch.
I'm not sure about disable TASKQ_DYNAMIC when `spl_taskq_thread_dynamic==0` though, we might want to be able to change that without reload the module. If you really dislike the `tqt_is_dynamic`, we can reset `ns_spawn` after `wait_event(tq->tq_wait_waitq, tq->tq_nthreads == count);`

tuxoko:
@dweeezil 
See if you like this one.
I also add another patch for taskq_wait_outstanding.

module/spl/spl-taskq.c
```diff
@@ -763,11 +763,12 @@ taskq_thread_spawn_task(void *arg)
 	taskq_t *tq = (taskq_t *)arg;
 	unsigned long flags;
 
-	(void) taskq_thread_create(tq);
-
-	spin_lock_irqsave_nested(&tq->tq_lock, flags, tq->tq_lock_class);
-	tq->tq_nspawn--;
-	spin_unlock_irqrestore(&tq->tq_lock, flags);
+	if (taskq_thread_create(tq) == NULL) {
+		/* restore spawning count if failed */
```
dweeezil: I wondered at first the context of this comment (when looking at the original patch) and thought "restore to _what_"?  How about:

    /* cancel the request to spawn if thread creation fails */

module/spl/spl-taskq.c
```diff
@@ -1063,6 +1072,8 @@ taskq_create(const char *name, int nthreads, pri_t pri,
 
 	/* Wait for all threads to be started before potential destroy */
 	wait_event(tq->tq_wait_waitq, tq->tq_nthreads == count);
+	/* reset spawning count to zero */
```
dweeezil: Maybe?:

    /* revert the changes to tq_nspawn adjustment in taskq_thread() */

dweeezil:
@tuxoko LGTM.

module/spl/spl-taskq.c
tuxoko: I don't see how that makes it more understandable. The creation is already failed, and we are not actively cancel anything.

module/spl/spl-taskq.c
tuxoko: This is more vague than saying "reset to zero", but I'll add the reason.

module/spl/spl-taskq.c
dweeezil: The original comment was just fine once I thought about it for a moment.

module/spl/spl-condvar.c
```diff
@@ -311,9 +311,8 @@ cv_timedwait_hires_common(kcondvar_t *cvp, kmutex_t *mp, hrtime_t tim, hrtime_t
 		tim = (tim / res) * res;
 	}
 
-	ASSERT(!(flag & CALLOUT_FLAG_ABSOLUTE));
-	/* get abs expire time */
-	tim += gethrtime();
+	if (!(flag & CALLOUT_FLAG_ABSOLUTE))
+		tim += gethrtime();
```
behlendorf: I'm not sure how I missed this when I looked over 39cd90e but yes we definitely need to restore this.

module/spl/spl-taskq.c
```diff
@@ -447,8 +447,8 @@ taskq_wait_outstanding_check(taskq_t *tq, taskqid_t id)
 void
 taskq_wait_outstanding(taskq_t *tq, taskqid_t id)
 {
-	wait_event(tq->tq_wait_waitq,
-	    taskq_wait_outstanding_check(tq, id ? id : tq->tq_next_id - 1));
+	id = id ? id : tq->tq_next_id - 1;
+	wait_event(tq->tq_wait_waitq, taskq_wait_outstanding_check(tq, id));
```
behlendorf: Good find!  I thought I'd observed this kind of behavior in the past and this explains it.

behlendorf:
@tuxoko these three LGTM.

e726590 Fix taskq_wait_outstanding re-evaluate tq_next_id
a815c53 Fix race between taskq_destroy and dynamic spawning thread
c1da2df Restore CALLOUT_FLAG_ABSOLUTE in cv_timedwait_hires

However, I'd like to hold off on this one until things settle down in the mainline kernel.

fc6269d Use schedule_timeout_interruptible for busy wait in splat-condvar.c

tuxoko:
I'm fine with that.

behlendorf:
Merged as:

b3a22a0 Fix taskq_wait_outstanding re-evaluate tq_next_id
5ce028b Fix race between taskq_destroy and dynamic spawning thread
872e0cc Restore CALLOUT_FLAG_ABSOLUTE in cv_timedwait_hires

PR #554, opened 2016-05-25
Implement a proper rw_tryupgrade

tuxoko:
Current rw_tryupgrade does rw_exit and then rw_tryenter(RW_RWITER), and then
does rw_enter(RW_READER) if it fails. This violate the assumption that
rw_tryupgrade should be atomic and could cause extra contention or even lock
inversion.

This patch we implement a proper rw_tryupgrade. For rwsem-spinlock, we take
the spinlock to check rwsem->count and rwsem->wait_list. For normal rwsem, we
use cmpxchg on rwsem->count to change the value from single reader to single
writer.

### DIFF
--- a/config/spl-build.m4
+++ b/config/spl-build.m4
@@ -39,6 +39,7 @@ AC_DEFUN([SPL_AC_CONFIG_KERNEL], [
 	SPL_AC_2ARGS_ZLIB_DEFLATE_WORKSPACESIZE
 	SPL_AC_SHRINK_CONTROL_STRUCT
 	SPL_AC_RWSEM_SPINLOCK_IS_RAW
+	SPL_AC_RWSEM_ACTIVITY
 	SPL_AC_SCHED_RT_HEADER
 	SPL_AC_2ARGS_VFS_GETATTR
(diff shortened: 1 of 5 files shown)


behlendorf:
@tuxoko could you open a dummy zfs PR so we can get a test run of this patch.  They'll give us some build coverage for a few architectures and kernels, plus some functional testing.

That said, this is a quite a bit smaller than I was expecting it to be which is a nice surprise.  Maybe this won't be as fragile as I expected.

dweeezil:
As I alluded to in zfsonlinux/zfs#4692 my only real concern was non-x86 architectures (ARM mainly) and where they are and may be going with their implementations.

tuxoko:
While it may seems that many archs have their own implementations. Actually, they are just optimization for the fast path. They all share the same structure and same slow path. So they should all be logically the same.

tuxoko:
update: fix splat use RWSEM_COUNT

behlendorf:
This LGTM.  I'm OK with supporting this approach at least until it's no longer maintainable for some reason.

PR #560, opened 2016-06-14
isa_defs.h should be more arch agnostic

5YN3R6Y:
I've found isa_defs.h to be somewhat of pain when compiling on anything other than x86. Many times it defines things that it shouldn't (such as _ILP32 when gcc already defines _LP64), especially when the compiler already had everything it needed pre-defined.

Now isa_defs.h respects these macros if they are already defined. Let me know if you think this needs anything changed. It's been tested on sparc64, amd64, and ppc64 using various different versions of gcc.

If clang (or any other compiler people use) doesn't work with this i can come up with something else.

### DIFF
--- a/include/sys/isa_defs.h
+++ b/include/sys/isa_defs.h
@@ -25,175 +25,69 @@
 #ifndef	_SPL_ISA_DEFS_H
 #define	_SPL_ISA_DEFS_H
 
-/* x86_64 arch specific defines */
-#if defined(__x86_64) || defined(__x86_64__)
+/*****************************************************************************\
+ *  If the compiler is already defining _LP64 or _ILP32 (but not both), we
+ *  should trust the compiler.
+\*****************************************************************************/
 
-#if !defined(__x86_64)
-#define __x86_64
-#endif
-
-#if !defined(__amd64)
-#define __amd64
-#endif
-
-#if !defined(__x86)
-#define __x86
-#endif
-
-#if !defined(_LP64)
-#define _LP64
-#endif
-
-/* i386 arch specific defines */
-#elif defined(__i386) || defined(__i386__)
-
-#if !defined(__i386)
-#define __i386
-#endif
-
-#if !defined(__x86)
-#define __x86
-#endif
-
-#if !defined(_ILP32)
-#define _ILP32
-#endif
-
-/* powerpc (ppc64) arch specific defines */
-#elif defined(__powerpc) || defined(__powerpc__) || defined(__powerpc64__)
-
-#if !defined(__powerpc)
-#define __powerpc
-#endif
-
-#if !defined(__powerpc__)
-#define __powerpc__
-#endif
-
-#if defined(__powerpc64__)
-#if !defined(_LP64)
-#define _LP64
-#endif
-#else
-#if !defined(_ILP32)
-#define _ILP32
-#endif
-#endif
-
-/* arm arch specific defines */
-#elif defined(__arm) || defined(__arm__) || defined(__aarch64__)
-
-#if !defined(__arm)
-#define __arm
-#endif
-
-#if !defined(__arm__)
-#define __arm__
-#endif
-
-#if defined(__aarch64__)
-#if !defined(_LP64)
-#define _LP64
-#endif
-#else
-#if !defined(_ILP32)
-#define _ILP32
-#endif
-#endif
-
-#if defined(__ARMEL__) || defined(__AARCH64EL__)
-#define _LITTLE_ENDIAN
-#else
-#define _BIG_ENDIAN
-#endif
-
-/* sparc arch specific defines */
-#elif defined(__sparc) || defined(__sparc__)
-
-#if !defined(__sparc)
-#define __sparc
-#endif
-
-#if !defined(__sparc__)
-#define __sparc__
-#endif
-
-#if defined(__arch64__)
-#if !defined(_LP64)
-#define _LP64
-#endif
-#else
-#if !defined(_ILP32)
-#define _ILP32
-#endif
-#endif
-
-#define _BIG_ENDIAN
-#define _SUNOS_VTOC_16
-
-/* s390 arch specific defines */
-#elif defined(__s390__)
-#if defined(__s390x__)
-#if !defined(_LP64)
-#define	_LP64
-#endif
-#else
-#if !defined(_ILP32)
-#define	_ILP32
-#endif
-#endif
-
-#define	_BIG_ENDIAN
-
-/* MIPS arch specific defines */
-#elif defined(__mips__)
-
-#if defined(__MIPSEB__)
-#define	_BIG_ENDIAN
-#elif defined(__MIPSEL__)
-#define	_LITTLE_ENDIAN
-#else
-#error MIPS no endian specified
-#endif
-
-#ifndef _LP64
-#define	_ILP32
-#endif
-
-#define	_SUNOS_VTOC_16
-
-#else
-/*
- * Currently supported:
- * x86_64, i386, arm, powerpc, s390, sparc, and mips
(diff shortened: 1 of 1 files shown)


behlendorf:
In principle I don't mind simplifying this but I'm not sure it's worth it.  Getting it right for all supported architectures, compilers, and kernels will take a fair bit of work.  As a starting point I've opened a dummy PR zfsonlinux/zfs#4761 to run the automated testing for this patch.  That resulted in quite a few build failures which need to investigated and resolved.  We'd also want to make an analogous change to the `isa_defs.h` in zfs.

ryao:
@behlendorf According to @F4S4K4N, this makes SPARC64 work. I am equally hesitant to simplify this for the reasons that you stated.

tonyhutter:
FYI,  this change will conflict with one of my pull requests (https://github.com/zfsonlinux/spl/pull/561/commits/17d00ace366f0d7c92b5c3f9de81a97beb79124e).

I like how our current isa_defs.h roughly matches Illumos's isa_defs.h in structure, especially in regard to the explicit architecture-specific sections (I.e. `#elif defined(__i386) || defined(__i386__) ...`).  It made it easy to identify where to put the  `_ALIGNMENT_REQUIRED` macros I needed to port over for the checksum code.  I worry that if we diverge too much from Illumos's version of the header, it will make it harder to port over future patches.  That said, I do think you have good reasons for wanting to clean up the header.

F4S4K4N:
@behlendorf @ryao 

I see what's going on with the build errors, let me re-work this patch a bit. If other compilers are a concern we can pull this information from headers instead.

@tonyhutter 

In theory this doesn't stop anyone from doing arch-specific things. It just takes the 32/64bit and endian defines out of being arch specific. Mostly because they aren't arch specific.

````#elif defined(__i386) || defined(__i386__) ...)

can still be added if you need to add arch specific macros, the only reason i removed those is because there is not arch specific code there currently.

@ryao 

Sparc64 support doesn't require this, I suppose i just saw it not making much sense to make if statements if we aren't doing anything arch specific. You also can run into weird edge cases where the arch specific flags aren't defined. SPARC is an example for this as

````#elif defined(__sparc) || defined(__sparc__)

doesn't always work for every supported version of gcc.

F4S4K4N:
Please see the updated pull request, it should build fine now. I've downloaded and compiled the latest spl and zfs on all three archs at my disposal to make sure it works correctly.

@tonyhutter this keeps all arch specific if sections open should anyone need to add arch specific macros.

Another point i forgot to make is that these changes should allow unsupported ISA's to build without failing on this header. Endianess and 32/64 bit will be determined on any arch. If the build fails it will be because of an actual issue, not because a header is preventing the build.

ryao:
@F4S4K4N There is a repository of information on what macros are defined for what architectures here:

https://sourceforge.net/p/predef/wiki/Architectures/

There is also information on endianness:

https://sourceforge.net/p/predef/wiki/Endianness/

If we can lean on the toolchain to do this for us, we should do it, but it is very important to validate that this behaves in a manner that is universally predictable and is not specific to any toolchain. Although I can think of a few simplifications, there will be a significant amount of work to gain the level of certainty that the simplification does not introduce regressions over the status quo and we ought to discuss what that would be before anyone spends any time on it. Here are my thoughts.

We must support any toolchain supported by Linux. 99% of the time, that is GCC and there are many supported versions of GCC. A google search suggests that the oldest supported version of GCC is 3.2:

http://www.linuxtopia.org/online_books/linux_kernel/kernel_configuration/ch03.html

Checking against every version is impractical, but we will need to check that things build and run properly on a decent sample of architectures. For the reason, I think we should check on each architecture against 3.4.6, GCC 4.2.4 and GCC 4.9.3. GCC 3.4.6 is the last of the GCC 3 series and should be representative of behavior for it. GCC 4.2.4 is the last release of the GCC 4.2 branch, which was the last GPLv2 branch and is one that should represent early GCC 4 well. GCC 4.9.3 is a more recent release and should represent the recent releases well.

At the bare minimum, we should build cross compilers and write some sort of fake input file to get out the preprocessor settings that matter for the sake of verifying sanity. Gentoo's crossdev tool can be used to simplify the task of building cross compilers that allow this. You would install crossdev, setup a local overlay, run commands like `crossdev x86_64-unknown-linux-gnu` as root and let it build the toolchain, which you can then access via commands like x86_64-unknown-linux-uclibc-gcc. The GCC version can be specified with `--gcc $ver` and versions from here:

https://packages.gentoo.org/packages/sys-devel/gcc

If you do not run Gentoo, you can just grab a stage3 tarball, extract it to $LOCATION (your choice), edit $LOCATION/etc/resolv.conf, and use it as a chroot. I have a script that can simplify this, the stage3 tarballs are easy to download and documentation on crossdev and overlays is readily accessible:

https://gist.github.com/ryao/3c345f206b19c9795109
http://distfiles.gentoo.org/releases/amd64/autobuilds/current-stage3-amd64/
https://wiki.gentoo.org/wiki/Overlay/Local_overlay
https://wiki.gentoo.org/wiki/Embedded_Handbook/General/Creating_a_cross-compiler

The current architectures by their GCC tuple names that are known to build with the present code are powerpc, powerpc64, i386, aarch64, arm, armeb, s390x and x86_64. Since the purpose of this is to add sparc support, we could use sparc and sparc64 too for completeness.

If you are willing to volunteer to check all of these after you, myself and @behlendorf agree on a methodology, that would be awesome and it could enable us to simplify this. However, it is easier to just extend the status quo.

F4S4K4N:
@ryao 

I use Gentoo and crossdev on almost a daily basis as it turns out. Ill do some some testing and post back when I have some better information.

behlendorf:
@F4S4K4N refreshed the buildbot results.  The patch fails on arm, ppc, and x86.  For example check out the [build failure for x86](http://build.zfsonlinux.org/builders/Ubuntu%2014.04%20i686%20%28BUILD%29/builds/5476/steps/shell_2/logs/make)

F4S4K4N:
@behlendorf are these all 32 bit variants of those archs?

behlendorf:
@F4S4K4N yes.

F4S4K4N:
@behlendorf __arch32__ seems to be a fairly new thing in gcc, i've been able to verify it's missing running ubuntu 14.04 in a VM. The updated pull request should fix that. I did a test build that passes on 14.04.

I've been doing some work with crossdev. i have passing builds on ppc64 (verified on a power6 machine), sparc64 (verified on various sun boxes), aarch64 (verified on an RPi3), and s390x (not tested).

Next ill grab some older gcc versions and do some more testing.

5YN3R6Y:
Sorry for the delay, getting everything setup has taken some time. I have tested the following scenarios on Gentoo linux via crossdev as well as some physical machines.

GCC 3.4.6 / 4.2.4 / 4.9.3
ppc - pass - tested on a g4 mac mini
ppc64 - pass - tested on an IBM JS21
i386 - pass - not tested
aarch64 - pass - tested on a RPi 3
arm - pass - tested on a RPi 2
armeb - pass - not physically tested
s390x - pass - not physically tested
x86_64 - pass - test on a dell poweredge R710
sparc - pass - not physically tested
sparc64 - pass - tested on a SunFire V215

While i did test on some physical machines, i did not do any extensive performance testing. I simply created a pool and did a few reads and writes to the pool.

It looks like i need to modify the patch to address some conflicts with the current tree, i will have a new one up shortly.

behlendorf:
@5YN3R6Y thank you for looking at this but giving this some more thought I don't think we should rework this code to be more arch agnostic.  At the moment it aligns well with the illumos version which is convenient, we know the existing version works well, and adding new architectures is currently straight forward.  Let's leave it as is.  Thanks.

Issue #563, opened 2016-07-26
Linux 4.8 compat


behlendorf:
New kernel release window, new build failures.

```
spl/spl/spl-rwlock.c: In function ‘__rwsem_tryupgrade’:
spl/spl/spl-rwlock.c:55:2: error: invalid initializer
  val = cmpxchg(&rwsem->count, SPL_RWSEM_SINGLE_READER_VALUE,
  ^
spl/spl/spl-rwlock.c:55:2: error: invalid initializer
spl/spl/spl-rwlock.c:57:14: error: invalid operands to binary == (have ‘atomic_long_t’ and ‘long int’)
  return (val == SPL_RWSEM_SINGLE_READER_VALUE);
              ^
spl/spl/spl-rwlock.c:58:1: warning: control reaches end of non-void function [-Wreturn-type]
 }
 ^
  CC      mm/rmap.o
make[2]: *** [spl/spl/spl-rwlock.o] Error 1
make[1]: *** [spl/spl] Error 2
make: *** [spl] Error 2
make: *** Waiting for unfinished jobs....
```

PR #564, opened 2016-07-26
Linux 4.8 compat: rw_semaphore atomic_long_t count

behlendorf:
For non-rwsem-spinlocks the "count" member was changed from a
"long" to "atomic_long_t" type.  A configure check has been
added to detect this change along with new versions of the
_rwsem_tryupgrade() function and RWSEM_COUNT() macro.  See
torvalds/linux/commit/8ee62b18 for complete details.

Issue #563

### DIFF
--- a/include/linux/rwsem_compat.h
+++ b/include/linux/rwsem_compat.h
@@ -35,9 +35,12 @@
 #define	SPL_RWSEM_SINGLE_WRITER_VALUE	(RWSEM_ACTIVE_WRITE_BIAS)
 #endif
 
-/* Linux 3.16 change activity to count for rwsem-spinlock */
-#ifdef HAVE_RWSEM_ACTIVITY
+/* Linux 3.16 changed activity to count for rwsem-spinlock */
+#if defined(HAVE_RWSEM_ACTIVITY)
 #define	RWSEM_COUNT(sem)	sem->activity
+/* Linux 4.8 changed count to an atomic_long_t for !rwsem-spinlock */
+#elif defined(HAVE_RWSEM_ATOMIC_LONG_COUNT)
+#define	RWSEM_COUNT(sem)	atomic_long_read(&(sem)->count)
 #else
 #define	RWSEM_COUNT(sem)	sem->count
 #endif
--- a/module/spl/spl-rwlock.c
+++ b/module/spl/spl-rwlock.c
@@ -32,7 +32,7 @@
 
 #define DEBUG_SUBSYSTEM S_RWLOCK
 
-#ifdef CONFIG_RWSEM_GENERIC_SPINLOCK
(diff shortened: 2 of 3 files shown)


behlendorf:
@tuxoko @dweeezil thus far this appears to be the only compatibility issue with the latest kernel.  If you guys could review this we can resolve the new kernel.org buildbot failures.  You'll probably want to wait for the buildbot test results in zfsonlinux/zfs#4887 even though I did test this locally.

behlendorf:
Looks like a spoke too soon, there's at least [one other](http://52.37.62.191/builders/Kernel.org%20Default%20x86_64%20%28BUILD%29/builds/5998/steps/shell_3/logs/make) build issue to resolve.

include/linux/rwsem_compat.h
```diff
@@ -35,12 +35,19 @@
...
 
-/* Linux 3.16 change activity to count for rwsem-spinlock */
+/*
+ * Linux 3.16 changed activity to count for rwsem-spinlock
+ * Linux 4.8 changed count to an atomic_long_t for !rwsem-spinlock
+ */
 #ifdef HAVE_RWSEM_ACTIVITY
 #define	RWSEM_COUNT(sem)	sem->activity
 #else
+#ifdef HAVE_RWSEM_ATOMIC_LONG_COUNT
```
tuxoko: `#elif defined()`

module/spl/spl-rwlock.c
```diff
@@ -48,6 +48,16 @@ __rwsem_tryupgrade(struct rw_semaphore *rwsem)
 	return (ret);
 }
 #else
+#ifdef HAVE_RWSEM_ATOMIC_LONG_COUNT
```
tuxoko: `#elif defined()`

tuxoko:
submit_bio change
torvalds/linux@4e49ea4a3d276365bf7396c9b77b4d1d5923835a

include/linux/rwsem_compat.h
behlendorf: Yeah, I went back and forth about doing this as an `#elif defined()`.  I'll refresh the patch with that change since after looking at the whole patch I think I like that better too.

behlendorf:
> submit_bio change

While I'll all set up for this I'll work up a patch.  It looks straight forward, just update `vdev_submit_bio` to OR in the bits to `bio->bi_rw`.

dweeezil:
@behlendorf These all (including the SPL change) LGTM at first glance.  I'll be testing them out on a current master kernel shortly.

behlendorf:
Merged as:

b7c7008 Linux 4.8 compat: rw_semaphore atomic_long_t count

PR #565, opened 2016-07-29
Trim unused sym fix

lorddoskias:


behlendorf:
Presumably most, maybe all, distribution kernels don't have CONFIG_TRIM_UNUSED_KSYM defined or this would have been observed sooner.  So adding this check shouldn't cause problems, correct?

lorddoskias:
I tested the checks with CONFIG_TRIM_UNUSED_KSYM set and not set. And it worked as expected - in the case where it wasn't set the building went fine. So I think it's safe. I've got the idea from the DEBUG_LOCK checks in zfs.

PR #566, opened 2016-08-01
Fix HAVE_MUTEX_OWNER test for kernels prior to 4.6

dweeezil:
Recent 4.X kernels prior to 4.6 require #include of spinlock.h in
order to get the definition of __ARCH_SPIN_LOCK_UNLOCKED which is used
by DEFINE_MUTEX().

[EDIT: missing space in commit comment fixed but not yet committed/pushed]


dweeezil:
Sorry, but I've not figured out which kernel commit prior to 4.6 obviates the need for ```#include <linux/spinlock.h``` nor whether this is a problem in any 3.X kernels.

behlendorf:
Interesting find, well even if we can't pin it down to a specific commit adding the additional include is safe and probably a good idea.

PR #567, opened 2016-08-19
Fix: handle NULL case in spl_kmem_free_track()

GeLiXin:
Reference issue: #https://github.com/zfsonlinux/zfs/issues/4967

When DEBUG_KMEM_TRACKING is enabled in SPL, we keep tracking all the buffers alloced by
kmem_alloc() and kmem_zalloc().  If a NULL pointer which indicates no track info in SPL
is passed to spl_kmem_free_track, we just ignore it.

### DIFF
--- a/module/spl/spl-kmem.c
+++ b/module/spl/spl-kmem.c
@@ -383,6 +383,10 @@ spl_kmem_free_track(const void *ptr, size_t size)
 {
 	kmem_debug_t *dptr;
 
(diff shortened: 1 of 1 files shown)


behlendorf:
LGTM

behlendorf:
Mergerd as:

aeb9baa Fix: handle NULL case in spl_kmem_free_track()

PR #569, opened 2016-09-13
Clean up in cred.h

tuxoko:
Remove the code that doesn't make any sense.


behlendorf:
Can you open a dummy PR for ZFS with a ` Requires-spl:` line just to make sure this doesn't unexpectedly break something.

PR #571, opened 2016-09-16
Increase spl_kmem_alloc_warn limit

behlendorf:
In order to support ABD with large blocks the spl_kmem_alloc_warn
limits needs to be increased to 64K.

A 16M block requires that a pointers be stored for 4096 4K-pages
on an x86_64 system.  Each of these points is 8 bytes requiring
an allocation of 8*4096=32,768 bytes.  The addition of a small
header to this structure pushes the allocation over the default
32K warning threshold.

In addition, fix a small bug when MAX was used instead of MIN
when setting the default.  This ensures a reasonable limit is
still set on systems with page sizes larger then 4K.

### DIFF
--- a/module/spl/spl-kmem.c
+++ b/module/spl/spl-kmem.c
@@ -35,7 +35,7 @@
  * rate limited warning will be printed to the console for any kmem_alloc()
  * which exceeds a reasonable threshold.
  *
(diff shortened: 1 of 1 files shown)


behlendorf:
@dpquigl can you review this so we can get it merged for your ABD patches.

PR #578, opened 2016-10-07
Fix out-of-bound in per_cpu in spl_random_init

tuxoko:
When iterating per_cpu values, we need to use for_each_cpu. While NR_CPUS
indicates the number of CPU supported by the kernel, it might not initialize
all of them if the kernel decides it's not possible to use them.

### DIFF
--- a/module/spl/spl-generic.c
+++ b/module/spl/spl-generic.c
@@ -638,7 +638,7 @@ spl_random_init(void)
 		    "0x%016llx%016llx.", cpu_to_be64(s[0]), cpu_to_be64(s[1]));
 	}
 
-	for (i = 0; i < NR_CPUS; i++) {
+	for_each_possible_cpu(i) {
 		uint64_t *wordp = per_cpu(spl_pseudo_entropy, i);
 
 		spl_rand_jump(s);


tuxoko:
@behlendorf 
This fix the bug you found when DEBUG_PAGEALLOC.

module/spl/spl-generic.c
```diff
@@ -638,7 +638,7 @@ spl_random_init(void)
 		    "0x%016llx%016llx.", cpu_to_be64(s[0]), cpu_to_be64(s[1]));
 	}
 
-	for (i = 0; i < NR_CPUS; i++) {
+	for_each_cpu(i, cpu_possible_mask) {
```
behlendorf: Let's use the preferred helper macro `for_each_possible_cpu` for this.

tuxoko: Right, didn't notice this.

tuxoko:
If by global you mean the `_GLOBAL__sub_I_65535_2_proc_spl_kstat` in your stack trace, I think that's just a misinterpret of the stack. I just pinpoint the crash happens in spl_random_init and the only thing caught my eye is the per_cpu thing.

tuxoko:
Actually, your stack trace looks pretty different than mine, it might be different issues after all. But it would be strange because I can run fine with this patch.

behlendorf:
OK, I see.  I thought I might be overlooking something in the stack trace.

PR #586, opened 2016-11-10
Limit number of tasks shown in taskq proc

tuxoko:
To prevent holding tq_lock for too long.

Before zfsonlinux/zfs@8e71ab9, hogging delay tasks and cat /proc/spl/taskq
would easily cause a lockup. While that bug has been fixed. It's probably
still a good idea to do this just in case task lists grow too large.

### DIFF
--- a/module/spl/spl-proc.c
+++ b/module/spl/spl-proc.c
@@ -234,6 +234,10 @@ taskq_seq_show_headers(struct seq_file *f)
 #define LHEAD_ACTIVE	4
 #define LHEAD_SIZE	5
 
+static unsigned int spl_max_show_tasks = 512;
+module_param(spl_max_show_tasks, uint, 0644);
+MODULE_PARM_DESC(spl_max_show_tasks, "Max number of tasks shown in taskq proc");
+
 static int
 taskq_seq_show_impl(struct seq_file *f, void *p, boolean_t allflag)
 {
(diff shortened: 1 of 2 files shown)


module/spl/spl-proc.c
```diff
@@ -308,16 +310,19 @@ taskq_seq_show_impl(struct seq_file *f, void *p, boolean_t allflag)
...
 				if (i == LHEAD_WAIT) {
 					wq = list_entry(lh, wait_queue_t, task_list);
 					if (j == 0)
 						seq_printf(f, "\t%s:",
 						    list_names[i]);
-					else if (j == 12) {
+					else if (j % 8 == 0)
 						seq_printf(f, "\n\t     ");
-						j = 0;
-					}
```
behlendorf: It look like the correct functioning of this loop depend on `j` being reset to be 0.  You'd need to modify the `if (j == 0)` conditional as well.

module/spl/spl-proc.c
```diff
@@ -234,6 +234,8 @@ taskq_seq_show_headers(struct seq_file *f)
 #define LHEAD_ACTIVE	4
 #define LHEAD_SIZE	5
 
+#define	MAX_SHOW_TASK	256
```
behlendorf: Perhaps this should be a tunable.  Often when I'm looking at the taskq kstat I'm debugging an issue and I want to make sure I'm seeing everything.  In these case it's entirely acceptable to stall the system to get the needed debugging.

tuxoko: Done.

dweeezil:
I was thinking this type of limit would be a good idea recently when debugging something which caused thousands of pending tasks (forgot what I was looking into).  I was going to suggest making the limit a tunable but I see the latest version of the patch already does so.

Any type of cap, however, prevents knowing the actual size of the list.  How about adding a counter to keep track of how large the list is and then printing out the size in the case of truncation?

PR #587, opened 2016-11-26
mutex: serialize exit()/destroy() without spinlock

ironMann:
Attempt to remove spinlock operations in mutex_exit().  Instead, perform synchronization in mutex_destroy() to avoid use after free.

### DIFF
--- a/config/spl-build.m4
+++ b/config/spl-build.m4
@@ -20,6 +20,7 @@ AC_DEFUN([SPL_AC_CONFIG_KERNEL], [
 	SPL_AC_DEBUG
 	SPL_AC_DEBUG_KMEM
 	SPL_AC_DEBUG_KMEM_TRACKING
+	SPL_AC_DEBUG_LOCK_TRACKING
 	SPL_AC_TEST_MODULE
 	SPL_AC_ATOMIC_SPINLOCK
 	SPL_AC_SHRINKER_CALLBACK
@@ -155,7 +156,7 @@ AC_DEFUN([SPL_AC_KERNEL], [
 
 	if test "$utsrelease"; then
 		kernsrcver=`(echo "#include <$utsrelease>";
-		             echo "kernsrcver=UTS_RELEASE") | 
+		             echo "kernsrcver=UTS_RELEASE") |
 		             cpp -I $kernelbuild/include |
 		             grep "^kernsrcver=" | cut -d \" -f 2`
 
@@ -226,7 +227,7 @@ AC_DEFUN([SPL_AC_RPM], [
 		AC_MSG_RESULT([$HAVE_RPMBUILD])
 	])
 
-	RPM_DEFINE_COMMON='--define "$(DEBUG_SPL) 1" --define "$(DEBUG_KMEM) 1" --define "$(DEBUG_KMEM_TRACKING) 1"'
+	RPM_DEFINE_COMMON='--define "$(DEBUG_SPL) 1" --define "$(DEBUG_KMEM) 1" --define "$(DEBUG_KMEM_TRACKING) 1" --define "$(DEBUG_LOCK_TRACKING) 1"'
 	RPM_DEFINE_UTIL=
 	RPM_DEFINE_KMOD='--define "kernels $(LINUX_VERSION)"'
 	RPM_DEFINE_DKMS=
@@ -516,6 +517,33 @@ AC_DEFUN([SPL_AC_DEBUG_KMEM_TRACKING], [
 	AC_SUBST(DEBUG_KMEM_TRACKING)
 	AC_MSG_CHECKING([whether detailed kmem tracking is enabled])
 	AC_MSG_RESULT([$enable_debug_kmem_tracking])
+])
+
+dnl #
+dnl # Disabled by default it provides detailed lock liveness tracking.
+dnl # This feature tracks location of init() calls for kmutex_t, kcondvar_t and
+dnl # krwlock_t, and matching destroy calls. List of unmatched variables is
+dnl # kprint-ed on SPL module unload.
+dnl #
+AC_DEFUN([SPL_AC_DEBUG_LOCK_TRACKING], [
+	AC_ARG_ENABLE([debug-lock-tracking],
+		[AS_HELP_STRING([--enable-debug-lock-tracking],
+		[Enable detailed lock init/destroy tracking  @<:@default=no@:>@])],
+		[],
+		[enable_debug_lock_tracking=no])
+
+	AS_IF([test "x$enable_debug_lock_tracking" = xyes],
+	[
+		DEBUG_LOCK_TRACKING="_with_debug_lock_tracking"
+		AC_DEFINE([DEBUG_LOCK_TRACKING], [1],
+		[Define to 1 to enable detailed lock tracking])
+	], [
+		DEBUG_LOCK_TRACKING="_without_debug_lock_tracking"
+	])
+
+	AC_SUBST(DEBUG_LOCK_TRACKING)
+	AC_MSG_CHECKING([whether detailed lock init/destroy tracking is enabled])
+	AC_MSG_RESULT([$enable_debug_lock_tracking])
 ])
 
 dnl #
--- a/include/spl-lockdbg.h
+++ b/include/spl-lockdbg.h
@@ -1,0 +1,114 @@
+/*
+ *  Copyright (C) 2016 Gvozden Neskovic <neskovic@gmail.com>
+ *
+ *  This file is part of the SPL, Solaris Porting Layer.
+ *  For details, see <http://zfsonlinux.org/>.
+ *
+ *  The SPL is free software; you can redistribute it and/or modify it
+ *  under the terms of the GNU General Public License as published by the
+ *  Free Software Foundation; either version 2 of the License, or (at your
+ *  option) any later version.
+ *
+ *  The SPL is distributed in the hope that it will be useful, but WITHOUT
+ *  ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ *  FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ *  for more details.
+ *
+ *  You should have received a copy of the GNU General Public License along
+ *  with the SPL.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef _SPL_LOCKDBG_H
+#define	_SPL_LOCKDBG_H
+
+#ifdef DEBUG_LOCK_TRACKING
+
+#include <linux/types.h>
+#include <linux/rbtree.h>
+#include <linux/hash.h>
+#include <linux/slab.h>
+#include <linux/ratelimit.h>
+#include <linux/stacktrace.h>
+#include <linux/seq_file.h>
+#include <linux/proc_fs.h>
+#include <linux/jiffies.h>
+
+#define	SLT_BCKT_SIZE_BITS	(9)
+#define	SLT_BCKT_SIZE		(1U << SLT_BCKT_SIZE_BITS)
+#define	SLT_HASH(x)		(hash_ptr((void *)x, SLT_BCKT_SIZE_BITS))
+
+typedef struct spl_lock_tracking {
+	rwlock_t		slt_rwlock;		/* global rwlock */
+	struct rb_root		slt_buckets[SLT_BCKT_SIZE]; /* tracked */
+	spinlock_t		slt_locks[SLT_BCKT_SIZE]; /* tracked locks */
+	atomic_t		slt_cnt;		/* tracked count */
+	struct rb_root		slt_mismatched;		/* mismatched */
+	spinlock_t		slt_mismatched_lock;	/* mismatched locks */
+	atomic_t		slt_mismatched_cnt;	/* mismatched count */
+} spl_lock_tracking_t;
+
+extern spl_lock_tracking_t lock_tracker;
+
+#define	SLT_FILE_LOC(l)		(strrchr(l, '/') ? strrchr(l, '/') + 1 : (l))
+
+#define	SLT_RWLOCK(sltp)		(&(sltp)->slt_rwlock)
+#define	SLT_BUCKET(sltp, idx)		(&((sltp)->slt_buckets[idx]))
+#define	SLT_BUCKET_LOCK(sltp, idx)	(&((sltp)->slt_locks[idx]))
+
+#define	SLT_TRACKED_NUM(sltp)		atomic_read(&((sltp)->slt_cnt))
+#define	SLT_TRACKED_INC(sltp)		atomic_inc(&((sltp)->slt_cnt))
+#define	SLT_TRACKED_DEC(sltp)		atomic_dec(&((sltp)->slt_cnt))
+
+#define	SLT_MISMATCHED(sltp)		(&((sltp)->slt_mismatched))
+#define	SLT_MISMATCHED_LOCK(sltp)	(&((sltp)->slt_mismatched_lock))
+#define	SLT_MISMATCHED_NUM(sltp)	atomic_read(&(sltp)->slt_mismatched_cnt)
+#define	SLT_MISMATCHED_INC(sltp)	atomic_inc(&(sltp)->slt_mismatched_cnt)
+#define	SLT_MISMATCHED_DEC(sltp)	atomic_dec(&(sltp)->slt_mismatched_cnt)
+
+#define	_OBJ_INIT_LOC1(x) #x
+#define	_OBJ_INIT_LOC2(x) _OBJ_INIT_LOC1(x)
+#define	OBJ_INIT_LOC __FILE__ ":" _OBJ_INIT_LOC2(__LINE__)
+
+typedef enum  lock_type {
+	SLT_MUTEX	= 0,
+	SLT_RWLOCK	= 1,
+	SLT_CONDVAR	= 2,
+} lock_type_t;
+
+typedef struct lock_tracking_node {
+	struct rb_node	lt_rbnode;
+	ulong_t		lt_ptr;
+	lock_type_t	lt_type;
+#define	LT_LOC_SIZE	32
+	char		lt_loc[LT_LOC_SIZE];
+#define	LT_STACK_TRACE_SIZE	22
+	ulong_t		lt_trace[LT_STACK_TRACE_SIZE];
+	ulong_t		lt_trace_nr;
+} lock_tracking_node_t;
+
+
+extern struct file_operations proc_rwlock_tracking_operations;
+extern struct file_operations proc_mutex_tracking_operations;
+extern struct file_operations proc_condvar_tracking_operations;
+
+/*
+ * Tracker interface:
+ * - spl_lock_tracking_record_init(lock_type_t, ulong_t, char *)
+ * - spl_lock_tracking_record_destroy(lock_type_t, ulong_t, char *)
+ */
+void spl_lock_tracking_record_init(lock_type_t type, ulong_t lp, char *loc);
+void spl_lock_tracking_record_destroy(lock_type_t type, ulong_t lp, char *loc);
+
+void spl_lockdbg_fini(void);
+
+#else
+
+#define	DEFINE_LOCK_TRACKING(module)	void *module = NULL
+
+#define	spl_lock_tracking_record_init(a, b, c)		do {} while (0)
+#define	spl_lock_tracking_record_destroy(a, b, c)	do {} while (0)
+#define	spl_lockdbg_fini()				do {} while (0)
+
+#endif /* DEBUG_LOCK_TRACKING */
+
+#endif /* _SPL_LOCKDBG_H */
--- a/include/sys/mutex.h
+++ b/include/sys/mutex.h
@@ -29,10 +29,12 @@
 #include <linux/mutex.h>
 #include <linux/compiler_compat.h>
 #include <linux/lockdep.h>
+#include <spl-lockdbg.h>
 
 typedef enum {
-	MUTEX_DEFAULT   = 0,
-	MUTEX_NOLOCKDEP = 1
+	MUTEX_DEFAULT    = 0,      /* Normal */
+	MUTEX_NOLOCKDEP  = 1 << 0, /* No LockDep */
+	MUTEX_NOTRACKING = 1 << 1, /* No Tracking */
 } kmutex_type_t;
 
 typedef struct {
@@ -43,7 +45,8 @@ typedef struct {
 
 /* Use upper portion of m_flags to store flags */
 #define	MUTEX_FLAG_NOLOCKDEP	((u32)1 << 31)
-#define	MUTEX_FLAG_MASK		(~(((u32)1 << 31) - 1))
+#define	MUTEX_FLAG_NOTRACKING	((u32)1 << 30)
+#define	MUTEX_FLAG_MASK		(~(((u32)1 << 30) - 1))
 
 #define	MUTEX(mp)		(&((mp)->m_mutex))
 #define	MUTEX_FLAGS(mp)		(atomic_read(&(mp)->m_flags) & MUTEX_FLAG_MASK)
@@ -67,14 +70,19 @@ spl_mutex_set_type(kmutex_t *mp, kmutex_type_t type)
 {
 	u32 t = 0;
 
+	ASSERT((type == MUTEX_DEFAULT) ||
+		((type & (MUTEX_NOLOCKDEP | MUTEX_NOTRACKING)) != 0));
 	/*
 	 * Mutex m_flags flag/counter bits:
 	 * bit 31 : NOLOCKDEP
-	 * bits [30 - 0]: mutex_exit counter (m_flags & ~MUTEX_FLAG_MASK)
+	 * bit 30 : NODEBUG
+	 * bits [29 - 0]: mutex_exit counter (m_flags & ~MUTEX_FLAG_MASK)
 	 */
-
-	if (type == MUTEX_NOLOCKDEP)
+	if (type & MUTEX_NOLOCKDEP)
 		t |= MUTEX_FLAG_NOLOCKDEP;
+
+	if (type & MUTEX_NOTRACKING)
+		t |= MUTEX_FLAG_NOTRACKING;
 
 	atomic_set(&mp->m_flags, t & MUTEX_FLAG_MASK);
 }
@@ -82,10 +90,20 @@ spl_mutex_set_type(kmutex_t *mp, kmutex_type_t type)
 static inline kmutex_type_t
 spl_mutex_get_type(kmutex_t *mp)
 {
-	if (MUTEX_FLAGS(mp) & MUTEX_FLAG_NOLOCKDEP)
-		return MUTEX_NOLOCKDEP;
+	kmutex_type_t type = MUTEX_DEFAULT;
+	u32 flags = 0;
 
-	return MUTEX_DEFAULT;
+	ASSERT3P(mp, !=, NULL);
+
+	flags = MUTEX_FLAGS(mp);
+
+	if (flags & MUTEX_FLAG_NOLOCKDEP)
+		type |= MUTEX_NOLOCKDEP;
+
+	if (flags & MUTEX_FLAG_NOTRACKING)
+		type |= MUTEX_NOTRACKING;
+
+	return type;
 }
 
 #define	mutex_owner(mp)		(ACCESS_ONCE((mp)->m_owner))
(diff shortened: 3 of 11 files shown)


include/sys/mutex.h
```diff
@@ -176,19 +178,22 @@ spl_mutex_lockdep_on_maybe(kmutex_t *mp)			\
...
-	spl_mutex_clear_owner(mp);				\
-	mutex_unlock(MUTEX(mp));				\
-	spin_unlock(&(mp)->m_lock);				\
-	spl_mutex_lockdep_on_maybe(mp);				\
+
+#undef mutex_destroy
+#define	mutex_destroy(mp)						\
+{									\
+	for(;;)	{							\
+		VERIFY3P(mutex_owner(mp), ==, NULL);			\
```
behlendorf: Without the spin lock it's possible to hit this VERIFY since you're deferring the serialization from `mutex_exit()` to `mutex_destroy()`.  Probably the right thing to do here is simply remove the VERIFY.  Then this `#define` could be restructured as:

```diff
#undef mutex_destroy
#define	mutex_destroy(mp)						\
{									\
	while (MUTEX_EXIT_CNT(mp) != 0)					\
		schedule_timeout_interruptible(1);			\￼
}
```

This change also makes it unsafe to call `mutex_destroy()` in a context which can't schedule.  Now I don't believe it's ever used that way so it's not a problem, but it is something to be aware of since it's a slightly change in behavior.

behlendorf (changes requested):
So this change effectively resolves the use-after-free issue by switching to a reference counting model in `mutex_destroy()` rather than a spin lock which is serialized in `mutex_exit()`.  This should be safe as long as we're not missing any calls to `mutex_destroy()` which are now going to be critical.

@tuxoko you authored the original change here can you comment on this.

tuxoko:
@behlendorf 
I'm fine with this as long as you think mutex_destroy is always used and not a problem.

include/sys/mutex.h
ironMann: Fixed. Hitting that VERIFY anywhere in `mutex_destroy()` would be a bug (destroying a locked mutex). Removal of the spinlock does not change semantics of this condition, the `owner` is protected by the real mutex.

behlendorf (approved):
LGTM, thanks for the quick refresh.

@tuxoko it should be used everywhere.  We found some instances where it was missing upstream and added the missing calls for Linux.

include/sys/mutex.h
```diff
@@ -176,19 +204,21 @@ spl_mutex_lockdep_on_maybe(kmutex_t *mp)			\
...
-	spl_mutex_clear_owner(mp);				\
-	mutex_unlock(MUTEX(mp));				\
-	spin_unlock(&(mp)->m_lock);				\
-	spl_mutex_lockdep_on_maybe(mp);				\
+
+#undef mutex_destroy
+#define	mutex_destroy(mp)						\
+{									\
+	for(;;)	{							\
+		if (MUTEX_COUNTER_VAL(mp) == 0)				\
```
tuxoko: `while (MUTEX_COUNTER_VAL(mp) != 0)`

tuxoko (changes requested):
I think we can add a debug option to use kmemleak to check for missing mutex_destroy.

include/sys/mutex.h
ironMann: Fixed.

ironMann:
@tuxoko @behlendorf  I'll try adding kmemleak hooks to see if those are useful, and if they yield many false positives.
Do you think we need a configure flag or enable it on all supported kernels if it works well?

behlendorf:
I think it'd be reasonable to enable it on all supported kernels built with kmemleak support.  No need for an option.

tuxoko:
@ironMann 
Yeah, just check CONFIG_DEBUG_KMEMLEAK. You can kmalloc a small buffer in mutex_init and kfree in mutex_destroy. That way if the mutex is freed without destroyed, it will get caught by kmemleak. You should also put the kmalloc inside a function, so the stack trace will be obvious.

ironMann:
@behlendorf @tuxoko I tried what you're suggesting, but kmemleak proved to be extremely unreliable. First I had to rebuild my debug kernel (fedora 25) because kmemleak was getting disabled during boot. Next, the kmemleak `scan` operation would return anywhere between 5 and 705 possible leaks. And I'm not completely sure if kmemleak will scan static kmutex_t objects.
I also tried using LOCKDEP, by having an extra mutex locked in init() and unlocked in destroy(), to have reports of freeing of a locked mutex. But that also reported huge amount of false positives...

I was able to track down (in my limited testing) about 375 missing invocations of `mutex_destroy()` using following one-liner:
```bash
find . -type f -name "*.c" | xargs grep -w "mutex_init\|mutex_destroy" | tr '(' ' ' | awk '{tot[$2"\t"$1]++} END { for (i in tot) print tot[i],"\t",i }' | sort -k3 -k2
```
Now the number of `mutex_init()` matches the number of `mutex_destroy()` calls: zfsonlinux/zfs#5428

In the end, do you think this is worth including currently?

tuxoko:
@ironMann 
It's the stuff you see related to mutex or unrelated?

ironMann:
@tuxoko Mostly unrelated, and transient. Lot of rcu, kswapd, and random syscall stacks. I was able to find 1 missing call in `zfs_create_fs()`. Ultimately, I didn't find it very useful.

behlendorf:
It sounds like we're not going to get too much value out of hooking this in to kmemleak.  In the interests of keeping things simple I'd suggest we not bother.  But we should fix the one missing `mutex_destroy()` in `zfs_create_fs()` as long as we're now aware of it.

tuxoko:
@ironMann @behlendorf 
I actually find it very useful. I can catch `zfs_ratelimit_init`, `dmu_objset_open_impl` and `zfs_create_fs`. So basically all the stuff in zfsonlinux/zfs#5428 except for icp. But I don't know how to exercise icp paths anyway.

ironMann:
The whole kmemleak infrastructure will disable itself if misused, or if its internal buffer is overflown (the size is a kernel compile option), and we can end up only leaking memory. The solution is just not elegant enough, IMO. We could keep counters of init()/destroy() innovations for debugging and warn on mismatch when spl unloads. zfsonlinux/zfs#5428 should account for most of the difference.

tuxoko:
@ironMann 
This is for debug purpose, why would leaking memory be a problem?
If it can catch missing mutex_destroy reliably, in my case it is. Then, we should definitely add it. Especially it seems that illumos doesn't have strict requirement for mutex_destroy.

ironMann:
@tuxoko @behlendorf IMO, the kmemleak is an unwieldy solution for the problem. The worst of all, it doesn't force the issue on you unless you know where to look and have the kernel configured for that. As always, it's just my opinion.

In the second commit, I added a tracing methods that will do precisely what we need, and report issues on SPL unload (enabled only in debug build). Comments welcome. Sample:

```
[ 1438.149941] *****************************************************
               * SPL Mutex Tracker: Missing mutex_destroy() calls: *
               *****************************************************
[ 1438.149995] [0xffff9ea89ff879c0] name/loc = dmu_objset.c:527
[ 1438.150016] [0xffff9ea8a086ed30] name/loc = zfs_comutil.c:225
[ 1438.150039] [0xffff9ea8a086edf8] name/loc = zfs_comutil.c:225
...
[ 1438.150308] [0xffff9ea8c1d299c0] name/loc = dmu_objset.c:527
[ 1438.150946] [0xffff9ea8c5d10cd8] name/loc = kcf_sched.c:1273
[ 1438.151569] [0xffff9ea8c5d10e60] name/loc = kcf_sched.c:1276
[ 1438.152197] [0xffff9ea8c5fb00c8] name/loc = kcf_sched.c:1109
[ 1438.152831] [0xffff9ea8c7f4c6d8] name/loc = zfs_znode.c:1843
[ 1438.153453] [0xffff9ea8c8a10000] name/loc = kcf_sched.c:1119
[ 1438.154108] [0xffff9ea8c8a12000] name/loc = kcf_sched.c:1119
...
[ 1438.164345] [0xffff9ea8d886a810] name/loc = kcf_prov_tabs.c:350
[ 1438.167266] [0xffff9ea90fb9d9c0] name/loc = dmu_objset.c:527
[ 1438.167994] [0xffff9ea91978c9c0] name/loc = dmu_objset.c:527
[ 1438.168718] [0xffff9ea91978e9c0] name/loc = dmu_objset.c:527
[ 1438.169446] [0xffffffffc08957e0] name/loc = kcf_sched.c:1127
[ 1438.170183] [0xffffffffc08958c0] name/loc = kcf_prov_tabs.c:82
[ 1438.170900] [0xffffffffc0895a60] name/loc = kcf_mech_tabs.c:161
[ 1438.171614] [0xffffffffc0895b48] name/loc = kcf_mech_tabs.c:242
[ 1438.172332] [0xffffffffc0895c40] name/loc = kcf_mech_tabs.c:242
...
```

tuxoko:
@ironMann 
I don't understand why you are so against kmemleak. It's a great tool that catches real memleak (https://github.com/zfsonlinux/zfs/issues/5244, https://github.com/zfsonlinux/zfs/issues/4707, https://github.com/zfsonlinux/zfs/issues/4706). And we really need to run it anyway at least for major releases. It might have some gatcha if you've never run it before, like the early log buffer, but it's not a big deal, it's not like every developer needs to run it.

Anyway here's my patch.

https://gist.github.com/tuxoko/0e508573c60a12f12bdf4b8215bac4fb

```
unreferenced object 0xffff8934c7a5ede0 (size 16):
  comm "mount.zfs", pid 5317, jiffies 4295124762 (age 778.692s)
  hex dump (first 16 bytes):
    26 6f 73 2d 3e 6f 73 5f 75 70 67 72 61 64 65 5f  &os->os_upgrade_
  backtrace:
    [<ffffffffac82d8ca>] kmemleak_alloc+0x4a/0xa0
    [<ffffffffac1ffd55>] kmem_cache_alloc_trace+0xf5/0x1d0
    [<ffffffffc06591b8>] spl_mutex_leak_init+0x28/0x50 [spl]
    [<ffffffffc097fb38>] dmu_objset_open_impl+0x3f8/0x880 [zfs]
    [<ffffffffc0980077>] dmu_objset_from_ds+0xb7/0x120 [zfs]
    [<ffffffffc098015f>] dmu_objset_hold+0x7f/0xb0 [zfs]
    [<ffffffffc09fc932>] zfs_ioc_objset_zplprops+0x32/0x160 [zfs]
    [<ffffffffc09fd51f>] zfsdev_ioctl+0x48f/0x520 [zfs]
    [<ffffffffac23d861>] do_vfs_ioctl+0xa1/0x5d0
    [<ffffffffac23de09>] SyS_ioctl+0x79/0x90
    [<ffffffffac838236>] entry_SYSCALL_64_fastpath+0x1e/0xa8
    [<ffffffffffffffff>] 0xffffffffffffffff
```

tuxoko:
Oh, BTW, I was able to catch icp stuff. It's just I need to rmmod.

rmmod makes the symbol name disappear. This is a minor annoyance, but you can still see the lock name. Or you can put file name and line number in the buffer like you did in your patch.

```
unreferenced object 0xffff893564b28480 (size 16):
  comm "modprobe", pid 5199, jiffies 4295118476 (age 1608.936s)
  hex dump (first 16 bytes):
    26 6b 63 66 70 6f 6f 6c 2d 3e 6b 70 5f 74 68 72  &kcfpool->kp_thr
  backtrace:
    [<ffffffffac82d8ca>] kmemleak_alloc+0x4a/0xa0
    [<ffffffffac1ffd55>] kmem_cache_alloc_trace+0xf5/0x1d0
    [<ffffffffc06591b8>] 0xffffffffc06591b8
    [<ffffffffc08261fe>] 0xffffffffc08261fe
    [<ffffffffc073e018>] 0xffffffffc073e018
    [<ffffffffac002190>] do_one_initcall+0x50/0x180
    [<ffffffffac194d44>] do_init_module+0x5f/0x1f6
    [<ffffffffac10d49d>] load_module+0x24dd/0x2b50
    [<ffffffffac10dd6f>] SYSC_finit_module+0xdf/0x110
    [<ffffffffac10ddbe>] SyS_finit_module+0xe/0x10
    [<ffffffffac838236>] entry_SYSCALL_64_fastpath+0x1e/0xa8
    [<ffffffffffffffff>] 0xffffffffffffffff
unreferenced object 0xffff893564b28490 (size 16):
  comm "modprobe", pid 5199, jiffies 4295118476 (age 1608.936s)
  hex dump (first 16 bytes):
    26 6b 63 66 70 6f 6f 6c 2d 3e 6b 70 5f 75 73 65  &kcfpool->kp_use
  backtrace:
    [<ffffffffac82d8ca>] kmemleak_alloc+0x4a/0xa0
    [<ffffffffac1ffd55>] kmem_cache_alloc_trace+0xf5/0x1d0
    [<ffffffffc06591b8>] 0xffffffffc06591b8
    [<ffffffffc0826259>] 0xffffffffc0826259
    [<ffffffffc073e018>] 0xffffffffc073e018
    [<ffffffffac002190>] do_one_initcall+0x50/0x180
    [<ffffffffac194d44>] do_init_module+0x5f/0x1f6
    [<ffffffffac10d49d>] load_module+0x24dd/0x2b50
    [<ffffffffac10dd6f>] SYSC_finit_module+0xdf/0x110
    [<ffffffffac10ddbe>] SyS_finit_module+0xe/0x10
    [<ffffffffac838236>] entry_SYSCALL_64_fastpath+0x1e/0xa8
    [<ffffffffffffffff>] 0xffffffffffffffff
unreferenced object 0xffff893564b284a0 (size 16):
  comm "modprobe", pid 5199, jiffies 4295118476 (age 1608.936s)
  hex dump (first 16 bytes):
    26 6e 74 66 79 5f 6c 69 73 74 5f 6c 6f 63 6b 00  &ntfy_list_lock.
  backtrace:
    [<ffffffffac82d8ca>] kmemleak_alloc+0x4a/0xa0
    [<ffffffffac1ffd55>] kmem_cache_alloc_trace+0xf5/0x1d0
    [<ffffffffc06591b8>] 0xffffffffc06591b8
    [<ffffffffc08262a9>] 0xffffffffc08262a9
    [<ffffffffc073e018>] 0xffffffffc073e018
    [<ffffffffac002190>] do_one_initcall+0x50/0x180
    [<ffffffffac194d44>] do_init_module+0x5f/0x1f6
    [<ffffffffac10d49d>] load_module+0x24dd/0x2b50
    [<ffffffffac10dd6f>] SYSC_finit_module+0xdf/0x110
    [<ffffffffac10ddbe>] SyS_finit_module+0xe/0x10
    [<ffffffffac838236>] entry_SYSCALL_64_fastpath+0x1e/0xa8
    [<ffffffffffffffff>] 0xffffffffffffffff
unreferenced object 0xffff893564b284d0 (size 16):
  comm "modprobe", pid 5199, jiffies 4295118476 (age 1608.936s)
  hex dump (first 16 bytes):
    26 64 65 73 63 2d 3e 70 64 5f 6c 6f 63 6b 00 00  &desc->pd_lock..
  backtrace:
    [<ffffffffac82d8ca>] kmemleak_alloc+0x4a/0xa0
    [<ffffffffac1ffd55>] kmem_cache_alloc_trace+0xf5/0x1d0
    [<ffffffffc06591b8>] 0xffffffffc06591b8
    [<ffffffffc0824b56>] 0xffffffffc0824b56
    [<ffffffffc0829295>] 0xffffffffc0829295
    [<ffffffffc082b4be>] 0xffffffffc082b4be
    [<ffffffffc073e01d>] 0xffffffffc073e01d
    [<ffffffffac002190>] do_one_initcall+0x50/0x180
    [<ffffffffac194d44>] do_init_module+0x5f/0x1f6
    [<ffffffffac10d49d>] load_module+0x24dd/0x2b50
    [<ffffffffac10dd6f>] SYSC_finit_module+0xdf/0x110
    [<ffffffffac10ddbe>] SyS_finit_module+0xe/0x10
    [<ffffffffac838236>] entry_SYSCALL_64_fastpath+0x1e/0xa8
    [<ffffffffffffffff>] 0xffffffffffffffff
```

ironMann:
@tuxoko 
Yes, I admit I'm probably irritated by the fact that I needed significantly more time to
get kmemleak working on a clean VM with a major distro, than to write some code to do precisely what I wanted to achieve. (and why does recompiling of fedora 25 kernel requires more that 20GiB of disk space?) I am also aware that throwing more code on the problem isn’t ideal, either. 
Kmemleak is a great tool, but it’s a necessary-evil type of tools, and as such it’s great that it exists. I’m just against masquerading other problems into something they’re not.

> It might have some gatcha if you've never run it before, like the early log buffer, but it's not a big deal, it's not like every developer needs to run it.

This is not how I perceived this specific problem. My understanding is that every patch, especially ones coming for other OSes should be checked for this issue. And the check should probably be extended to other spl primitives with init/destroy interface.

To conclude, let me summarize again my main concerns of using the kmemleak here:
- Unbounded memory leak. Get a init/destroy mismatch in just a right place and you'll get thousands of small memory leaks. If you’re a developer unaware that this is happening you’ll have a really hard time finding the cause. (could be mediated by limiting the number of living locks)
- False positives. I don’t know what is your experience, but number of reported leaks was varying greatly from scan to scan.
- Failing to report issues in a timely manner.
- Interfering with hunting for real memory leaks

I hope this makes my position more clear. [Last test run](http://build.zfsonlinux.org/builders/CentOS%207%20x86_64%20Mainline%20%28TEST%29/builds/614/steps/shell_11/logs/console) with my patch found one more candidate.

tuxoko:
> This is not how I perceived this specific problem. My understanding is that every patch, especially ones coming for other OSes should be checked for this issue. And the check should probably be extended to other spl primitives with init/destroy interface.

More reason to add kmemleak to the buildbot then.

module/spl/spl-mutex.c
```diff
@@ -24,13 +24,228 @@
...
+void
+spl_mutex_fini(void)
+{
+	spl_mutex_record_empty_and_report();
+}
+
+void
+spl_mutex_init_obj(kmutex_t *mp, char *name, kmutex_type_t type, void *arg)
+{
+	static struct lock_class_key spl_mutex_key;
```
tuxoko: This need to be in macro.

tuxoko (changes requested):

module/spl/spl-mutex.c
```diff
@@ -24,13 +24,228 @@
...
+	spin_lock_irqsave(&mtx_track_lock, flags);
+
+	mtp = spl_mutex_record_lookup((unsigned long)mp);
+	/* Check if exists */
+	if (mtp != NULL) {
+		printk(KERN_WARNING "SPL Mutex Tracker: "
+		    "double initialization of mutex: [0x%p] name/loc: %s\n"
+		    "was previously: [0x%p] name/loc: %s",
+		    (void *)mp, name, (void *)mtp->mt_ptr, mtp->mt_name);
+		goto out;
```
tuxoko: Maybe you should replace it so you don't loose track of new mutex.

behlendorf:
> irritated by the fact that I needed significantly more time to get kmemleak working on a clean VM with a major distro, than to write some code to do precisely what I wanted to achieve

This is absolutely something I can sympathize with.  Getting kmemleak running locally also took me far longer than I initially expected.  It's definitely a useful tool, but it's not something I'd want to ask other developers to have to setup regularly.  But as @tuxoko mentioned a one time cost of adding a buildbot builder is pretty reasonable.

Let's go with @ironMann's patch for the moment for the reasons he mentioned and I happen to agree with.  I'll bump up on my to do list getting a kmemleak TEST builder added to the buildbot.  @tuxoko I'm not sure if you use ec2 but if you happen to have an AMI around I could use that would make it easy to add.

tuxoko:
@behlendorf 
I'm afraid I haven't used ec2 before. But I'll check it out.

ironMann:
Do you think I should try to make this more general so it can be applied elsewhere? condvar and rwlocks come to mind, since they also have init/destroy interface.

behlendorf:
@ironMann that would be good.  We already do something similar to this with the `--enable-debug-kmem-tracking` configure option for kmem leak tracing.  It would be nice if any leak tracing we added used the same core infrastructure.  I'd also make sure to add a configure option to enable/disable it.  Disabled by default and then we'll enable it for the buildbot.

behlendorf:
@ironMann what's you're plan for this PR?  I'm OK with merging it as if if you want to open a new PR to tackle making the leak handling more generic.

ironMann:
@behlendorf I'll need few more days to get the generic version through, and enable the support for other primitives. Currently, https://github.com/zfsonlinux/spl/pull/587/commits/f1f8a0c115e7f64f346715b36db65a23fd085ae1 can be merged together with zfsonlinux/zfs/pull/5428 which fixes problems detected during testing.

behlendorf:
@ironMann you're choice.  There's no rush, I'm happy to wait if you'd like to fold your additional debugging improvements in to this PR.  Or if you'd rather do this in two parts then just go ahead and update this and the counterpart ZFS PR so they can be reviewed one last time before being merged.

ironMann:
@behlendorf What do you think about printing trace results into `seq_file`(s) ? Something like `/proc/spl/debug/locks/{kmutex_t,kcondvar_t,krwlock_t}`
This would help reduce overhead of taking full stack traces, and prevent losing the symbol names when other modules unload before spl.

behlendorf:
@ironMann that sounds entirely reasonable to me.  @dweeezil used the same `seq_file()` interfaces to implement the `/proc/spl/taskq` and `/proc/spl/taskq-all`.  These could both be reasonably moved in to a `/proc/spl/debug/` directory.  Maybe you could provide example output for what you have in mind?

include/sys/rwlock.h
```diff
@@ -134,26 +134,40 @@ RW_LOCK_HELD(krwlock_t *rwp)
...
+#endif
+
 /*
  * The following functions must be a #define and not static inline.
  * This ensures that the native linux semaphore functions (down/up)
  * will be correctly located in the users code which is important
  * for the built in kernel lock analysis tools
  */
 #define rw_init(rwp, name, type, arg)					\
-({									\
```
tuxoko: Please, don't do this. You either `({ })` or `do { } while (0)`. There are old code that only use `{ }` but you shouldn't add any more, and definitely not reverting it.

tuxoko:
@ironMann 
Can you clarify what exactly you want to show in the proc file and why would it have anything to do with symbol name?

Also, wouldn't most of the output happen when unloading modules? How would we be able to use proc file? I thought originally you just store the name in buffer, why do you want to change that?

ironMann:
@behlendorf @tuxoko  For all locks that are missing destroy() call the output would consist of pair <file:line, stack trace> where the lock is initialized. All of this is currently printed at spl unload, sample [here](http://build.zfsonlinux.org/builders/Amazon%202015.09%20x86_64%20SIMD%20%28TEST%29/builds/2830/steps/shell_7/logs/console). 

Workflow could be:
 - Load add modules
 - Run your usual tests
 - Check `spl/debug/lock` files (can resolve all stack trace symbols at this stage)
 - Unload modules (possibly also printk findings on SPL unload)

behlendorf:
> For all locks that are missing destroy() 

Then I expect we'll see all the missing destroy() calls for long lived things which are initialized in `_init()` and cleaned up in `_fini()`.  This might make it somewhat difficult to just be able to eyeball the output to look for leaks.  Although I really like the idea of being able to easily look at a proc file on the running system  to get this information.

It might be prudent to add `MUTEX_NODEBUG`, `RW_NODEBUG`, `CV_NODEBUG`, types which could be used to suppress tracking for specific locks.  Not only would this be helpful for certain long lived locks we'd prefer not to track but there may be some locks we want to exclude due to excessive performance impact.  Alternately, we'll need to build this kind of logic in to the test scripts to look for only unexpected leaks.

ironMann:
> Then I expect we'll see all the missing destroy() calls for long lived things which are initialized in _init() and cleaned up in _fini(). This might make it somewhat difficult to just be able to eyeball the output to look for leaks. 

The patch currently tracks and reports two kinds of locks separately.
If a lock (a memory location) has been initialized 2 times without destroy (usually locks embedded in other structures), then it can be considered as real positive, and location of such locks is reported only once. The unique key here is `file:line`. Also reported is the stack trace when the first init() is performed. These are good candidates to print into proc files.

The global tracking tree still contains all long lived locks and they can not be positively resolved until all modules unload (I call them "missing teardown" locks), as the tracker lives in SPL.

tuxoko:
I don't see much benefits of stack trace. A lock either is in an object or a global lock, so the missing destroy is either in the object free path or *_fini(). If the lock name along cannot easily let us figure out where the destroy is missing, then there's something wrong with the code readability and it need to be fixed.

So my point is we should make it simple and just stick to the original printing file, line, and lock name. And don't bother with the proc file and stack trace.

include/spl-lockdbg.h
```diff
@@ -0,0 +1,297 @@
...
+			kfree(mtp);
+		}
+		spin_unlock_irqrestore(SLT_BUCKET_LOCK(sltp, i), flags);
+	}
+
+
+}
+
+static void
+spl_lock_tracking_record_init(spl_lock_tracking_t *sltp, ulong_t lp, char *loc)
```
behlendorf: Rather than implement everything in this header lets move all of these functions in to `module/spl/spl-lockdbg.c`.  Then we can do the conventional thing and provide structures, defines, and public prototypes in the header.

config/spl-build.m4
```diff
@@ -524,9 +524,9 @@ dnl #
 AC_DEFUN([SPL_AC_DEBUG_LOCK_TRACKING], [
 	AC_ARG_ENABLE([debug-lock-tracking],
 		[AS_HELP_STRING([--enable-debug-lock-tracking],
-		[Enable detailed lock init/destroy tracking  @<:@default=no@:>@])],
+		[Enable detailed lock init/destroy tracking  @<:@default=yes@:>@])],
 		[],
-		[enable_debug_lock_tracking=no])
+		[enable_debug_lock_tracking=yes])
```
behlendorf: This can be accomplished by adding the following line to your top most commit comment.

```
CONFIG_OPTIONS="--enable-debug-lock-tracking --enable-debug"
```

behlendorf (changes requested):
@ironMann sorry about the delay reviewing this.  I've made a few inlines comments but if you'd like to revive this (and I think we should!), can you please.

* Rebase on the latest master and update the counterpart ZFS PR.  There are just a few trivial conflicts but it'll get us a fresh test run.

* The new functionality should be automatically enabled for testing since we build packages with `--enable-debug` and I saw you've extended that to imply `--enable-debug-lock-tracking` which is good.

* Manually run `cstyle -cpP` on all the new files you're adding and the ones you've significantly reworked.

include/sys/mutex.h
```diff
@@ -78,16 +80,24 @@ spl_mutex_set_type(kmutex_t *mp, kmutex_type_t type)
...
-	 * bits [30 - 0]: mutex_exit counter (m_flags & ~MUTEX_FLAG_MASK)
+	 * bit 30 : NODEBUG
+	 * bits [29 - 0]: mutex_exit counter (m_flags & ~MUTEX_FLAG_MASK)
 	 */
-
-	if (type == MUTEX_NOLOCKDEP)
-		t |= MUTEX_FLAG_NOLOCKDEP;
+	switch (type) {
+		case MUTEX_NODEBUG:
+			t |= MUTEX_FLAG_NODEBUG;
```
ironMann: Yes, I initially intended for NODEBUG to rule out all other checks.
Or should we add this check as a separate category, e.g. MUTEX_LOCKTRACKING?

include/spl-lockdbg.h
ironMann: will do. In that case I think we should track all types of primitives within a single instance of hash/rb_tree structure, it should scale well. All entries are matched only by the lock struct address, so we'll be able to catch if a lock is initialized over an exesting lock of different type.

ironMann:
@behlendorf Sorry for delays, work pivoted away from filesystems, so this is now a new hobby.
I addressed most of your comments above. 

few questions:
> The new functionality should be automatically enabled for testing... and I saw you've extended that to imply --enable-debug-lock-tracking which is good.

I'm not sure how to accomplish this. Adding `CONFIG_OPTIONS` didn't help (zfsonlinux/zfs#6107).
Testers should be extended to `cat /proc/spl/debug/*` files before any module is unloaded to get the most out of stack traces. Only file:line is printed during spl_fini.

behlendorf:
@ironMann sorry I dropped the ball on moving this one forward again, but we still do want these improvements!  As for a path forward now that 0.7 is tagged we have a little more flexibility regarding pulling in bigger changes.  We're also in progress of pulling the SPL source in to the ZFS source tree to simplify things.

If it's alright with you, why don't we hold off merging this a little longer until we pull the SPL in tree.  Then you can re-propose these changes in a PR against ZFS for review, testing, and merging.  Thanks for your patience!

ironMann:
@behlendorf No worries, just let me know what's the good time to restart the efforts.

behlendorf:
@ironMann that took a bit longer than expected, but the spl has been merged.  If you're still interested in this PR please go ahead and open a new one against the zfs repo.

PR #589, opened 2016-12-11
Add support for rw semaphore changes under PREEMPT_RT_FULL

clefru:
The main complication from the RT patch set is that the RW semaphore locks change in that read locks on an rwsem can be taken only by a single thread. All other threads are locked out. This single thread can take a read lock multiple times though. The underlying implementation changes to a mutex with an additional read_depth count.

The implementation can be best understood by inspecting the [RT patch](https://drive.google.com/open?id=0B1gmniVI7XGqNDUyRm8tN1JvNHc). rwsem_rt.h and rt.c give the best insight into how RT rwsem works. My implementation for rwsem_tryupgrade is basically an inversion of rt_downgrade_write found in rt.c. Please see the comments in the code.

Unfortunately, I have to drop SPLAT rwlock test4 completely as this test tries to take multiple locks from different threads, which RT rwsems do not support. Otherwise SPLAT, zconfig.sh, zpios-sanity.sh and zfs-tests.sh pass on my Debian-testing VM with linux-image-4.8.0-1-rt-amd64.

Assuming this PR is the right direction, I'll add a test to execute "rw_enter(rwp, RW_READER); rw_enter(rwp, RW_READER); ASSERT(rwsem_tryupgrade(rwp) == -EBUSY);" in the same thread.

### DIFF
--- a/include/sys/rwlock.h
+++ b/include/sys/rwlock.h
@@ -213,14 +213,6 @@ RW_LOCK_HELD(krwlock_t *rwp)
 	spl_rw_lockdep_on_maybe(rwp);					\
 })
 
-/*
- * This implementation of rw_tryupgrade() behaves slightly differently
- * from its counterparts on other platforms.  It drops the RW_READER lock
- * and then acquires the RW_WRITER lock leaving a small window where no
- * lock is held.  On other platforms the lock is never released during
- * the upgrade process.  This is necessary under Linux because the kernel
- * does not provide an upgrade function.
- */
 #define rw_tryupgrade(rwp)						\
 ({									\
 	int _rc_ = 0;							\
--- a/module/spl/spl-rwlock.c
+++ b/module/spl/spl-rwlock.c
@@ -32,7 +32,35 @@
 
 #define DEBUG_SUBSYSTEM S_RWLOCK
 
-#if defined(CONFIG_RWSEM_GENERIC_SPINLOCK)
+#if defined(CONFIG_PREEMPT_RT_FULL)
+
+#include <linux/rtmutex.h>
+
+static int
+__rwsem_tryupgrade(struct rw_semaphore *rwsem)
+{
+	ASSERT(rt_mutex_owner(&rwsem->lock) == current);
+
+	/* Under the realtime patch series, rwsem is implemented as a
+	 * single mutex held by readers and writers alike. However,
+	 * this implementation would prevent a thread from taking a
+	 * read lock twice, as the mutex would already be locked on
+	 * the second attempt. Therefore the implementation allows a
+	 * single thread to take a rwsem as read lock multiple times
+	 * tracking that nesting as read_depth counter.
+	 */
+	if (rwsem->read_depth <= 1) {
+		/* In case, the current thread has not taken the lock
+		 * more than once as read lock, we can allow an
+		 * upgrade to a write lock. rwsem_rt.h implements
+		 * write locks as read_depth == 0.
+		 */
+		rwsem->read_depth = 0;
+		return (1);
+	}
+	return (0);
+}
+#elif defined(CONFIG_RWSEM_GENERIC_SPINLOCK)
 static int
 __rwsem_tryupgrade(struct rw_semaphore *rwsem)
 {
--- a/module/splat/splat-rwlock.c
+++ b/module/splat/splat-rwlock.c
@@ -106,6 +106,15 @@ void splat_init_rw_priv(rw_priv_t *rwp, struct file *file)
 	rwp->rw_type = 0;
 }
 
+#if defined(CONFIG_PREEMPT_RT_FULL)
+static int
+splat_rwlock_test1(struct file *file, void *arg)
+{
+	// This test will never succeed on PREEMPT_RT_FULL because
+	// locks can only be held by a single thread.
+	return 0;
+}
+#else
 static int
 splat_rwlock_wr_thr(void *arg)
 {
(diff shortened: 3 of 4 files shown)


kernelOfTruth:
Testing right now with 4.8.14-rt9, runs fine so far, thanks :+1:

module/spl/spl-rwlock.c
```diff
@@ -32,7 +32,32 @@
...
 
-#if defined(CONFIG_RWSEM_GENERIC_SPINLOCK)
+#if defined(CONFIG_PREEMPT_RT_FULL)
+
+#include <linux/rtmutex.h>
+
+static int
+__rwsem_tryupgrade(struct rw_semaphore *rwsem)
+{
+	ASSERT(rt_mutex_owner(&rwsem->lock) != current);
```
behlendorf: Don't you mean `==`?  The lock can only be upgrades when your holding it as a reader.  Which would be exclusive in this case.  

```
ASSERT(rt_mutex_owner(&rwsem->lock) == current);
```

module/splat/splat-rwlock.c
```diff
@@ -207,6 +207,10 @@ splat_rwlock_test1(struct file *file, void *arg)
 	rw_thr_t rwt[SPLAT_RWLOCK_TEST_COUNT];
 	rw_priv_t *rwp;
 
+#if defined(CONFIG_PREEMPT_RT_FULL)
+	// This test will never succeed on PREEMPT_RT_FULL because locks can only be held by a single thread.
```
behlendorf: [cstyle] 80 character limit

module/splat/splat-rwlock.c
```diff
@@ -207,6 +207,10 @@ splat_rwlock_test1(struct file *file, void *arg)
...
+	return 0;
+#endif
```
behlendorf: I'd suggest turning this in to an #else clause for readability.

include/sys/rwlock.h
```diff
@@ -213,14 +213,6 @@ RW_LOCK_HELD(krwlock_t *rwp)
...
 })
 
-/*
- * This implementation of rw_tryupgrade() behaves slightly differently
- * from its counterparts on other platforms.  It drops the RW_READER lock
- * and then acquires the RW_WRITER lock leaving a small window where no
- * lock is held.  On other platforms the lock is never released during
- * the upgrade process.  This is necessary under Linux because the kernel
- * does not provide an upgrade function.
- */
```
clefru: I've split the commit in two.

module/spl/spl-rwlock.c
clefru: Yes, this is absolutely inverted. Fixed. The owner must be the current thread otherwise it's not legal to call the upgrade method. (Same condition as in the downgrade code).

ironMann: @clefru this puzzled me too, initially. But, wasn't the first statement right? You your'e doing upgrade, you are only a writer, hence not a owner.

clefru: I just reinspected 0124-rt-Add*.patch linked above. I am pretty sure the ASSERT is correct as is. Just search for rt_mt_owner in that file and you see a lot of comments that explain the semantic of that field, in particular that `rt_mutex_owner(lock) == current` when current owns the lock.

I also found why my code was wrong. My first try had the same guard `BUG_ON(rt_mutex_owner(&rwsem->lock) != current);` as the rt_downgrade_write method, but BUG_ON isn't available in spl, so I changed it to ASSERT(..) forgetting to invert the conditional.

clefru:
I haven't found a way to run cstyle quickly, so I just visually inspected the patch. Pardon me if I have not spotted a mistake.

clefru:
FYI on retesting the patch with zpios-sanity.sh, I hit an occasional deadlock. I haven't looked into that yet. 

[  726.461501] INFO: task zpios_io/0:3844 blocked for more than 120 seconds.
[  726.461525]       Tainted: P           OE   4.8.0-1-rt-amd64 #1
[  726.461558] "echo 0 > /proc/sys/kernel/hung_task_timeout_secs" disables this message.
[  726.461583] zpios_io/0      D ffff97f3bfce9100     0  3844      2 0x00000000
[  726.461587]  ffff97f3a6a1af40 ffff97f3ba7d8000 ffffffff83ea94b5 ffff97f3a62dc000
[  726.461591]  ffff97f3a6a1af40 ffff97f3ad48daa8 ffff97f3ad48dad0 ffff97f3a1ea6000
[  726.461594]  ffff97f3ad48da00 ffffffff8440db23 ffff97f3a6a1af40 ffff97f3a62dbec8
[  726.461598] Call Trace:
[  726.461601]  [<ffffffff83ea94b5>] ? preempt_count_add+0x5/0xa0
[  726.461604]  [<ffffffff8440db23>] ? schedule+0x43/0xd0
[  726.461606]  [<ffffffffc09b0df3>] ? zpios_thread_main+0x533/0xa20 [zpios]
[  726.461610]  [<ffffffffc09b08c0>] ? zpios_dmu_object_create+0x120/0x120 [zpios]
[  726.461613]  [<ffffffff83ea2bcd>] ? kthread+0xcd/0xf0
[  726.461616]  [<ffffffff844118ef>] ? ret_from_fork+0x1f/0x40
[  726.461618]  [<ffffffff83ea2b00>] ? kthread_worker_fn+0x150/0x150

module/spl/spl-rwlock.c
behlendorf: This is somewhat confusing because with the RT kernel there really is no such thing as a read lock.  They're all exclusive mutexs and thus have owners even when only held for read.  Offhand I can't think of any place in the ZFS code where this change in semantics will result in a problem but it's something we should keep in mind.

@clefru thanks for updating this, the BUG_ON -> ASSERT inversion makes perfect sense.

PR #590, opened 2016-12-15
Fix splat memleak and refactor

tuxoko:
SPLAT_TEST_FINI didn't call kfree causing memleak.

Also, refactor the code by making splat_test_{init,fini},
splat_subsystem_{init,fini} into functions.


module/splat/splat-ctl.c
```diff
@@ -33,9 +33,9 @@
  *  the kmem interfaces have been implemented correctly.  When the splat
  *  module is loaded splat_*_init() will be called for each subsystems
  *  tests.  It is the responsibility of splat_*_init() to register all
- *  the tests for this subsystem using the SPLAT_TEST_INIT() macro.
+ *  the tests for this subsystem using the splat_test_init() macro.
```
drescherjm: splat_test_init() macro. Should be function.

tuxoko:
@behlendorf 
> isn't that easy to spot

Right, I'll seperate it.

It's strange though the buildbot is still showing the leak on the first scan but not the second scan
http://build.zfsonlinux.org/builders/Ubuntu%2016.04%20x86_64%20Kmemleak%20%28TEST%29/builds/12
Do we start every test with fresh instance from image?

behlendorf:
> Do we start every test with fresh instance from image?

We start every buildbot build with a fresh instance from the image.  But not between individual test steps.

behlendorf:
Thanks for the quick refresh.  This works as expected for me locally, I'll get this merged and then we'll see where we're at with kmemleak.

tuxoko:
Oh I see, we use master version with zimport test, so the leaks shown is from master.

PR #595, opened 2017-01-09
Add support for recent kmem_cache_create_usercopy

kofee:
SLAB_USERCOPY flag was used to indicate PAX
    not to kill copies from kernel to userland.

    With recent grsecurity patchset and
    CONFIG_GRKERNSEC_HIDESYM that enables
    CONFIG_PAX_USERCOPY zfs would panic.

    Handle newer API while keeping old one
    functional.

Signed-off-by: Kevin Tanguy <kevin.tanguy@ovh.net>

### DIFF
--- a/module/spl/spl-kmem-cache.c
+++ b/module/spl/spl-kmem-cache.c
@@ -986,7 +986,9 @@ spl_kmem_cache_create(char *name, size_t size, size_t align,
 		if (rc)
 			goto out;
 	} else {
+#if !defined(KMEM_CACHE_USERCOPY)
 		unsigned long slabflags = 0;
+#endif
 
 		if (size > (SPL_MAX_KMEM_ORDER_NR_PAGES * PAGE_SIZE)) {
 			rc = EINVAL;
@@ -1001,8 +1003,17 @@ spl_kmem_cache_create(char *name, size_t size, size_t align,
 		slabflags |= SLAB_USERCOPY;
 #endif
 
-		skc->skc_linux_cache = kmem_cache_create(
-		    skc->skc_name, size, align, slabflags, NULL);
+#if defined(KMEM_CACHE_USERCOPY)
+        /*
+         * Newer grsec patchset uses kmem_cache_create_usercopy()
+         * instead of SLAB_USERCOPY flag
+         */
+        skc->skc_linux_cache = kmem_cache_create_usercopy(
+            skc->skc_name, size, 0, 0, 0, size, NULL);
+#else
+        skc->skc_linux_cache = kmem_cache_create(
+            skc->skc_name, size, align, slabflags, NULL);
+#endif
 		if (skc->skc_linux_cache == NULL) {
 			rc = ENOMEM;
 			goto out;
--- a/module/spl/spl-kmem-cache.c
+++ b/module/spl/spl-kmem-cache.c
@@ -986,9 +986,7 @@ spl_kmem_cache_create(char *name, size_t size, size_t align,
 		if (rc)
 			goto out;
 	} else {
-#if !defined(KMEM_CACHE_USERCOPY)
 		unsigned long slabflags = 0;
-#endif
 
 		if (size > (SPL_MAX_KMEM_ORDER_NR_PAGES * PAGE_SIZE)) {
 			rc = EINVAL;
@@ -1003,13 +1001,13 @@ spl_kmem_cache_create(char *name, size_t size, size_t align,
 		slabflags |= SLAB_USERCOPY;
 #endif
 
-#if defined(KMEM_CACHE_USERCOPY)
+#if defined(HAVE_KMEM_CACHE_CREATE_USERCOPY)
         /*
          * Newer grsec patchset uses kmem_cache_create_usercopy()
          * instead of SLAB_USERCOPY flag
          */
         skc->skc_linux_cache = kmem_cache_create_usercopy(
-            skc->skc_name, size, 0, 0, 0, size, NULL);
+            skc->skc_name, size, align, slabflags, 0, size, NULL);
 #else
         skc->skc_linux_cache = kmem_cache_create(
             skc->skc_name, size, align, slabflags, NULL);
--- a/config/spl-build.m4
+++ b/config/spl-build.m4
@@ -50,6 +50,7 @@ AC_DEFUN([SPL_AC_CONFIG_KERNEL], [
 	SPL_AC_MUTEX_OWNER
 	SPL_AC_INODE_LOCK
 	SPL_AC_GROUP_INFO_GID
+	SPL_AC_KMEM_CACHE_CREATE_USERCOPY
 ])
 
 AC_DEFUN([SPL_AC_MODULE_SYMVERS], [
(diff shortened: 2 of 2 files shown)


module/spl/spl-kmem-cache.c
```diff
@@ -1001,8 +1003,17 @@ spl_kmem_cache_create(char *name, size_t size, size_t align,
...
 
-		skc->skc_linux_cache = kmem_cache_create(
-		    skc->skc_name, size, align, slabflags, NULL);
+#if defined(KMEM_CACHE_USERCOPY)
+        /*
+         * Newer grsec patchset uses kmem_cache_create_usercopy()
+         * instead of SLAB_USERCOPY flag
+         */
+        skc->skc_linux_cache = kmem_cache_create_usercopy(
+            skc->skc_name, size, 0, 0, 0, size, NULL);
```
behlendorf: Why was 'align' dropped?  This needs to be preserved.

module/spl/spl-kmem-cache.c
```diff
@@ -986,7 +986,9 @@ spl_kmem_cache_create(char *name, size_t size, size_t align,
 		if (rc)
 			goto out;
 	} else {
+#if !defined(KMEM_CACHE_USERCOPY)
 		unsigned long slabflags = 0;
+#endif
```
behlendorf: Let's pass the `slabflags` to `kmem_cache_create_usercopy()` anyway even if it's always a zero and drop this new `#if` conditional.

module/spl/spl-kmem-cache.c
```diff
@@ -1001,8 +1003,17 @@ spl_kmem_cache_create(char *name, size_t size, size_t align,
...
 #endif
 
-		skc->skc_linux_cache = kmem_cache_create(
-		    skc->skc_name, size, align, slabflags, NULL);
+#if defined(KMEM_CACHE_USERCOPY)
+        /*
+         * Newer grsec patchset uses kmem_cache_create_usercopy()
+         * instead of SLAB_USERCOPY flag
+         */
+        skc->skc_linux_cache = kmem_cache_create_usercopy(
```
behlendorf: Rather than relying on KMEM_CACHE_USERCOPY being defined to imply `kmem_cache_create_usercopy()` exists a configure check should be added.  You can model the new check on the existing SPL_AC_FS_STRUCT_SPINLOCK check in `spl-build.m4`.  It's important to call the new function with local variables which match the types used by the target function.  This way API changes can be caught at configure time.

module/spl/spl-kmem-cache.c
kofee: done

module/spl/spl-kmem-cache.c
kofee: done too

module/spl/spl-kmem-cache.c
kofee: This was indeed a mistake on my side, corrected.

sempervictus:
I'm still hitting _invalid initializer_ errors when trying to build in-tree (in order to build for/with RAP). The patch does seem to allow building with HIDESYM out of tree, though RAP isn't an option at that point (or rather a more complicated option when dealing with packaging all of this).

Kernel config options i believe to be relevant:
```
CONFIG_PAX_NOEXEC=y
CONFIG_PAX_PAGEEXEC=y
CONFIG_PAX_EMUTRAMP=y
CONFIG_PAX_MPROTECT=y
CONFIG_PAX_KERNEXEC_PLUGIN=y
CONFIG_PAX_KERNEXEC_PLUGIN_METHOD_BTS=y
CONFIG_PAX_ASLR=y
CONFIG_PAX_RANDKSTACK=y
CONFIG_PAX_RANDUSTACK=y
CONFIG_PAX_RANDMMAP=y
CONFIG_PAX_MEMORY_SANITIZE=y
CONFIG_PAX_MEMORY_STACKLEAK=y
CONFIG_PAX_MEMORY_STRUCTLEAK=y
CONFIG_PAX_REFCOUNT=y
CONFIG_PAX_USERCOPY=y
CONFIG_PAX_SIZE_OVERFLOW=y
CONFIG_PAX_SIZE_OVERFLOW_EXTRA=y
CONFIG_PAX_INITIFY=y
CONFIG_HAVE_PAX_INITIFY_INIT_EXIT=y
CONFIG_PAX_LATENT_ENTROPY=y
CONFIG_PAX_RAP=y
CONFIG_GRKERNSEC_KMEM=y
CONFIG_GRKERNSEC_IO=y
CONFIG_GRKERNSEC_BPF_HARDEN=y
CONFIG_GRKERNSEC_PERF_HARDEN=y
CONFIG_GRKERNSEC_RAND_THREADSTACK=y
CONFIG_GRKERNSEC_PROC_MEMMAP=y
CONFIG_GRKERNSEC_KSTACKOVERFLOW=y
CONFIG_GRKERNSEC_BRUTE=y
CONFIG_GRKERNSEC_MODHARDEN=y
CONFIG_GRKERNSEC_HIDESYM=y
CONFIG_GRKERNSEC_RANDSTRUCT=y
CONFIG_GRKERNSEC_RANDSTRUCT_PERFORMANCE=y
```
Build error output:
```
spl/spl/spl-proc.c:631:3: error: invalid initializer
  {0},
   ^
spl/spl/spl-proc.c:631:3: note: (near initialization for ‘spl_kmem_table[6].<anonymous>’)
spl/spl/spl-proc.c:635:3: error: invalid initializer
  {0},
   ^
spl/spl/spl-proc.c:635:3: note: (near initialization for ‘spl_kstat_table[0].<anonymous>’)
  CC      mm/vmacache.o
spl/spl/spl-proc.c:666:11: error: invalid initializer
         { 0 },
           ^
spl/spl/spl-proc.c:666:11: note: (near initialization for ‘spl_table[4].<anonymous>’)
spl/spl/spl-proc.c:675:11: error: invalid initializer
         { 0 }
           ^
spl/spl/spl-proc.c:675:11: note: (near initialization for ‘spl_dir[1].<anonymous>’)
spl/spl/spl-proc.c:687:4: error: invalid initializer
  { 0 }
    ^
spl/spl/spl-proc.c:687:4: note: (near initialization for ‘spl_root[1].<anonymous>’)
make[2]: *** [scripts/Makefile.build:289: spl/spl/spl-proc.o] Error 1
make[1]: *** [scripts/Makefile.build:440: spl/spl] Error 2
```

Git blame says the code at L631 of spl-proc.c came in back in 2008... 

Instructions for build are taken from https://github.com/zfsonlinux/zfs/issues/851.
Is this one of those fun C89/C99 standards issues, or something in my actual config? Doesn't happen when building out of tree...

behlendorf:
> Is this one of those fun C89/C99 standards issues.

Yes, exactly.  This is unrelated to the original issue being addressed here but you should be able to resolve it by changing the `{ 0 }` to `{ }`.  Can verify this fix works, is so I'll open a PR for it.

```diff
diff --git a/module/spl/spl-proc.c b/module/spl/spl-proc.c
index 345d0d5..d6159b4 100644
--- a/module/spl/spl-proc.c
+++ b/module/spl/spl-proc.c
@@ -628,11 +628,11 @@ static struct ctl_table spl_kmem_table[] = {
                 .mode     = 0444,
                 .proc_handler = &proc_doslab,
         },
-       {0},
+       {},
 };
 
 static struct ctl_table spl_kstat_table[] = {
-       {0},
+       {},
 };
 
 static struct ctl_table spl_table[] = {
@@ -663,7 +663,7 @@ static struct ctl_table spl_table[] = {
                .mode     = 0555,
                .child    = spl_kstat_table,
        },
-        { 0 },
+        {},
 };
 
 static struct ctl_table spl_dir[] = {
@@ -672,7 +672,7 @@ static struct ctl_table spl_dir[] = {
                 .mode     = 0555,
                 .child    = spl_table,
         },
-        { 0 }
+        {}
 };
 
 static struct ctl_table spl_root[] = {
@@ -684,7 +684,7 @@ static struct ctl_table spl_root[] = {
        .mode = 0555,
        .child = spl_dir,
        },
-       { 0 }
+       {}
 };
 
 int
```

behlendorf:
@kofee thanks for the prompt update.  The updated patch LGTM.

kofee:
Thank you :)

tuxoko:
Can we use a KMC_USERCOPY flag to config this?

behlendorf:
That's a good thought, it's nicer than assuming that every kmem cache should be a usercopy cache and consistent with the intent of the patch.  I'd be OK with adding a KMC_USERCOPY flag which is only functional on kernels which support that functionality.

spendergrsec:
While PaX no longer supports SLAB_USERCOPY, grsec should still be maintaining support for it -- if it's not working, then I missed something, but you can check mm/slab_common.c and see the handling for it there.  It basically reroutes the code to the _usercopy variant of the slab cache creation, allowing copies to/from anywhere in the object.

Also, the struct initialization change was needed for RANDSTRUCT, as we randomize the layout of ctl_table.  The error arises because we require designated initializers for each struct member (otherwise non-designated initializers would end up initializing the wrong randomized fields).  We enforce this with a trick that causes the compilation error you see, by inserting a void struct member at the beginning of the struct, it will always generate a compile error when non-designated initializers are used.  Since it has no size, it disappears from the resulting structure layout and doesn't affect the code generated.

Both of these changes look fine to me though -- in the future feel free to email me directly (spender@grsecurity.net) and I'd be happy to help you guys with any questions you have about fixing compatibility.

-Brad

behlendorf:
@spendergrsec thanks for detailed explanation and addition review, it's appreciated.  If we need any help resolving future compatibly problems or have questions I'll definitely let you know.  Thanks for offering.

I'll be merging this patch as is.  We can consider further refinements at a latter data if someone is interested in working on them.

PR #603, opened 2017-02-23
Linux 4.11 compat: set_task_state() removed

ofaaland:
Replace uses of set_task_state(current, STATE) with
set_current_state(STATE).

In Linux 4.11, torvalds/linux@642fa44, set_task_state() is removed.

All spl uses are of the form set_task_state(current, STATE).
set_current_state(STATE) is equivalent and has been available since
Linux 2.2.26.

Furthermore, set_current_state(STATE) is already used in about 15
locations within spl.  This change should have no impact other than
removing an unnecessary dependency.


ofaaland:
See https://github.com/zfsonlinux/zfs/pull/5815 for the test results.

PR #605, opened 2017-02-23
Do not attempt to try to allocate forever in spl_kmem_alloc_impl

sargun:
Previously, if the system was under memory pressure, spl_kmem_alloc_impl
could loop forever attempting to allocate memory. This patch puts
an upper bound on the attempts.

Signed-off-by: Sargun Dhillon <sargun@sargun.me>

### DIFF
--- a/module/spl/spl-kmem.c
+++ b/module/spl/spl-kmem.c
@@ -30,6 +30,14 @@
 #include <linux/ratelimit.h>
 
 /*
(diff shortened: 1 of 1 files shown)


ironMann:
For compatibility reasons `spl_kmem_alloc_impl()` must succeed when called with `KM_SLEEP`. A lot of code depend on this. See the [man page](https://docs.oracle.com/cd/E53394_01/html/E54779/kmem-alloc-9f.html)

PR #608, opened 2017-03-07
Linux 4.11 compat: vfs_getattr() takes 4 args

ofaaland:
There are changes to vfs_getattr() in torvalds/linux@a528d35.  The new
interface is:

int vfs_getattr(const struct path *path, struct kstat *stat,
               u32 request_mask, unsigned int query_flags)

The request_mask argument indicates which field(s) the caller intends to
use.  Fields the caller does not specify via request_mask may be
returned anyway, but may be approximate.

The query_flags argument indicates whether the filesystem must update
the attributes from the backing store.

This patch uses the query_flags which result in vfs_getattr behaving the same
as it did with the 2-argument version which the kernel provided before
Linux 4.11.

Members blksize and blocks are now always the same size regardless of
arch.  They match the size of the equivalent members in vnode_t.

The configure checks are modified to ensure that the appropriate
vfs_getattr() interface is used.

A more complete fix, removing the ZFS dependency on vfs_getattr()
entirely, is deferred as it is a much larger project.

### DIFF
--- a/module/spl/spl-vnode.c
+++ b/module/spl/spl-vnode.c
@@ -153,7 +153,9 @@ vn_open(const char *path, uio_seg_t seg, int flags, int mode,
 	if (IS_ERR(fp))
 		return (-PTR_ERR(fp));
 
-#ifdef HAVE_2ARGS_VFS_GETATTR
+#if defined(HAVE_4ARGS_VFS_GETATTR)
+	rc = vfs_getattr(&fp->f_path, &stat, STATX_TYPE, AT_STATX_SYNC_AS_STAT);
+#elif defined(HAVE_2ARGS_VFS_GETATTR)
 	rc = vfs_getattr(&fp->f_path, &stat);
 #else
 	rc = vfs_getattr(fp->f_path.mnt, fp->f_dentry, &stat);
@@ -510,7 +512,10 @@ vn_getattr(vnode_t *vp, vattr_t *vap, int flags, void *x3, void *x4)
 
 	fp = vp->v_file;
 
-#ifdef HAVE_2ARGS_VFS_GETATTR
+#if defined(HAVE_4ARGS_VFS_GETATTR)
+	rc = vfs_getattr(&fp->f_path, &stat, STATX_BASIC_STATS,
+	    AT_STATX_SYNC_AS_STAT);
+#elif defined(HAVE_2ARGS_VFS_GETATTR)
 	rc = vfs_getattr(&fp->f_path, &stat);
 #else
 	rc = vfs_getattr(fp->f_path.mnt, fp->f_dentry, &stat);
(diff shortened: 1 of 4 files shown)


module/spl/spl-vnode.c
```diff
@@ -153,7 +153,9 @@ vn_open(const char *path, uio_seg_t seg, int flags, int mode,
 	if (IS_ERR(fp))
 		return (-PTR_ERR(fp));
 
-#ifdef HAVE_2ARGS_VFS_GETATTR
+#if defined(HAVE_4ARGS_VFS_GETATTR)
+	rc = vfs_getattr(&fp->f_path, &stat, STATX_MODE, AT_STATX_SYNC_AS_STAT);
```
tuxoko:
```
#define STATX_TYPE		0x00000001U	/* Want/got stx_mode & S_IFMT */
#define STATX_MODE		0x00000002U	/* Want/got stx_mode & ~S_IFMT */
```
I think you use the wrong one.

ofaaland: You're right, thanks for catching that.

ofaaland:
A question about the configure checks for the vfs_getattr() interface change.
As I structured it here, I put the newest interface test at the top - checking for 4 arguments first.  I did that because it makes it easy to remove the old one when the time comes, and hints at which interface is the newest.

However, it makes the diff quite a bit larger.  I could instead just add the 4-arg test in the deepest fail case.

Do you have an opinion?

tuxoko:
I think we should split each test as a function and do cascade call. Something like this https://github.com/zfsonlinux/zfs/commit/7a789346afec87a89c4cde465faf9be4996db556

It would be easier to read, and we don't have to keep indenting it. It would also diff well if we change them.

ofaaland:
@tuxoko , there have been some changes to the SPL code since your approval.  Please look again when you get a chance, thanks.

module/spl/spl-vnode.c
```diff
@@ -153,9 +153,11 @@ vn_open(const char *path, uio_seg_t seg, int flags, int mode,
...
 
-#ifdef HAVE_2ARGS_VFS_GETATTR
+#if defined(HAVE_4ARGS_VFS_GETATTR)
+	rc = vfs_getattr(&fp->f_path, &stat, STATX_TYPE, AT_STATX_SYNC_AS_STAT);
+#elif defined(HAVE_2ARGS_VFS_GETATTR)
 	rc = vfs_getattr(&fp->f_path, &stat);
-#else
+#elif defined(HAVE_3ARGS_VFS_GETATTR)
 	rc = vfs_getattr(fp->f_path.mnt, fp->f_dentry, &stat);
 #endif
```
tuxoko: If would be better to use `#else` for the last one so that you don't end up with nothing here. Or add a `#else #error`

ofaaland: Agreed.

PR #610, opened 2017-03-31
glibc 2.25 compat: remove assert(X=Y)

ofaaland:
The assert() related definitions in glibc 2.25 were altered to warn
about assert(X=Y) when -Wparentheses is used.  See
https://abi-laboratory.pro/tracker/changelog/glibc/2.25/log.html

lib/list.c used this construct to set the value of a magic field which
is defined only when debugging.

Replaced the assert()s with #ifndef/#endifs.

### DIFF
--- a/lib/list.c
+++ b/lib/list.c
@@ -220,7 +220,9 @@ list_create (ListDelF f)
     l->fDel = f;
     l->count = 0;
     list_mutex_init(&l->mutex);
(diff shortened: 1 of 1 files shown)


ofaaland:
This was detected building on arch, which as of about 2 weeks ago was using glibc 2.25.  The build fails due to -Wparentheses being promoted to an error.

Tested with {noformat}splat --all{noformat} both with debug and non-debug builds.

PR #613, opened 2017-04-30
Support 4.9.20-rt16 style RW semaphores.

clefru:
Thanks to Thomas Glexiner <tglx@linutronix.de> for feedback on the implementation.


clefru:
I just found that splat rwlock:all tests are flaky and lead to deadlocks when run a couple of times. Closing.

PR #616, opened 2017-05-24
Fix cv_timedwait timeout

behlendorf:
Perform the already past expiration time check before updating
cvp->cv_mutex with the provided mutex.  This check only depends
on local state, and doing it first ensures that cvp->cv_mutex
will not get updated in the timeout case.


behlendorf:
Will do.

PR #622, opened 2017-06-19
Fix RWSEM_SPINLOCK_IS_RAW check failed

tuxoko:
Initialize dummy_lock to fix the build error in gcc 7.1.1 with:
  error: ‘dummy_lock’ is used uninitialized in this function


tuxoko:
This need to be in 0.6.5.x as it would cause build failure on that branch.

tuxoko:
Wrapped the line.

PR #625, opened 2017-07-04
Prevent calls to dh_shlibdeps

aerusso:
This patch addresses zfsonlinux/zfs#6309 and is the companion to ZFS patch zfsonlinux/zfs#6106 .

This patch is not strictly necessary, because that issue only affects `.deb` packages built by the ZFS package. See the other pull request for additional details.


aerusso:
Per the discussion in zfsonlinux/zfs#6106, exclusions can be passed directly to `dh_shlibdeps`, rather than a wholesale avoiding of the call. Since no such exclusions are needed with the SPL packaging, no change is required. Please disregard this pull request.

PR #633, opened 2017-07-25
Remove misguided HAVE_MUTEX_OWNER check

### LINKED ISSUE #632: HAVE_MUTEX_OWNER checks are misjudged
I noticed that ZFS has this HAVE_MUTEX_OWNER check and then does stuff to internal in-kernel mutex structures.
The problem with that is it's unsafe, for example once you enable lockdep (which I see zfs is prepared to do with the DEBUG_LOCK_ALLOC check hoping that mutex_lock_nested will become a non-GPL exported symbol), kernel would use the owner field in mutex in incompatible ways, in particular once a tainted module is loaded (like zfs), the owner field will always be zero triggering mutex_destroy assertion on zfs right away like this:

[21597.305519] VERIFY3(((*(volatile typeof((&((&zo->zo_lock)->m_mutex))->owner) *)&((&((&zo->zo_lock)->m_mutex))->owner))) == ((void *)0)) failed (ffff88030be28500 == (null))
[21597.307259] PANIC at zfs_onexit.c:104:zfs_onexit_destroy()
[21597.307792] Showing stack for process 3626
[21597.308343] CPU: 0 PID: 3626 Comm: mkfs.lustre Tainted: P           OE  ------------   3.10.0-debug #1
[21597.309366] Hardware name: Red Hat KVM, BIOS 0.5.1 01/01/2011
[21597.309886]  ffffffffa046228f 0000000012e37b1d ffff880308c3bcc0 ffffffff816fd404
[21597.310969]  ffff880308c3bcd0 ffffffffa01d8754 ffff880308c3be58 ffffffffa01d881f
[21597.312000]  ffff880308c3bcf0 ffffffff00000030 ffff880308c3be68 ffff880308c3be08
[21597.312984] Call Trace:
[21597.313477]  [<ffffffff816fd404>] dump_stack+0x19/0x1b
[21597.313993]  [<ffffffffa01d8754>] spl_dumpstack+0x44/0x50 [spl]
[21597.314529]  [<ffffffffa01d881f>] spl_panic+0xbf/0xf0 [spl]
[21597.315096]  [<ffffffffa039c6e6>] ? zfs_onexit_destroy+0x126/0x280 [zfs]
[21597.315667]  [<ffffffffa039c73c>] zfs_onexit_destroy+0x17c/0x280 [zfs]
[21597.316238]  [<ffffffffa03955d8>] zfsdev_release+0x48/0xd0 [zfs]

Even if we patch out the taint-disables-lockdep check out in the kernel, this assertion persists, just not as immediate.

As such I believe HAVE_MUTEX_OWNER check needs to be removed and ZFS should retain its own owner field in its own structure for debugging at all times.

verygreen:
It is just plain unsafe to peek inside in-kernel
mutex structure and make assumptions about what kernel
does with those internal fields like owner.

Kernel is all too happy to stop doing the expected things
like tracing lock owner once you load a tainted module
like spl/zfs that is not GPL.
As such you will get instant assertion failures like this:

[21597.305519] VERIFY3(((*(volatile typeof((&((&zo->zo_lock)->m_mutex))->owner) *)&((&((&zo->zo_lock)->m_mutex))->owner))) == ((void *)0)) failed (ffff88030be28500 == (null))
[21597.307259] PANIC at zfs_onexit.c:104:zfs_onexit_destroy()
[21597.307792] Showing stack for process 3626
[21597.308343] CPU: 0 PID: 3626 Comm: mkfs.lustre Tainted: P OE ------------ 3.10.0-debug #1
[21597.309366] Hardware name: Red Hat KVM, BIOS 0.5.1 01/01/2011
[21597.309886] ffffffffa046228f 0000000012e37b1d ffff880308c3bcc0 ffffffff816fd404
[21597.310969] ffff880308c3bcd0 ffffffffa01d8754 ffff880308c3be58 ffffffffa01d881f
[21597.312000] ffff880308c3bcf0 ffffffff00000030 ffff880308c3be68 ffff880308c3be08
[21597.312984] Call Trace:
[21597.313477] [] dump_stack+0x19/0x1b
[21597.313993] [] spl_dumpstack+0x44/0x50 [spl]
[21597.314529] [] spl_panic+0xbf/0xf0 [spl]
[21597.315096] [] ? zfs_onexit_destroy+0x126/0x280 [zfs]
[21597.315667] [] zfs_onexit_destroy+0x17c/0x280 [zfs]
[21597.316238] [] zfsdev_release+0x48/0xd0 [zfs]

This fixes #632
Signed-off-by: Oleg Drokin <green@linuxhacker.ru>

### DIFF
--- a/include/sys/mutex.h
+++ b/include/sys/mutex.h
@@ -40,10 +40,8 @@ typedef enum {
 typedef struct {
 	struct mutex		m_mutex;
 	spinlock_t		m_lock;	/* used for serializing mutex_exit */
-#ifndef HAVE_MUTEX_OWNER
 	/* only when kernel doesn't have owner */
 	kthread_t		*m_owner;
-#endif
 #ifdef CONFIG_LOCKDEP
 	kmutex_type_t		m_type;
 #endif /* CONFIG_LOCKDEP */
@@ -58,24 +56,16 @@ spl_mutex_set_owner(kmutex_t *mp)
 	 * kernel will handle its owner, so we don't need to do anything if it
 	 * is defined.
 	 */
-#ifndef HAVE_MUTEX_OWNER
 	mp->m_owner = current;
-#endif
 }
 
 static inline void
 spl_mutex_clear_owner(kmutex_t *mp)
 {
-#ifndef HAVE_MUTEX_OWNER
 	mp->m_owner = NULL;
-#endif
 }
 
-#ifdef HAVE_MUTEX_OWNER
 #define	mutex_owner(mp)		(ACCESS_ONCE(MUTEX(mp)->owner))
-#else
-#define	mutex_owner(mp)		(ACCESS_ONCE((mp)->m_owner))
-#endif
 #define	mutex_owned(mp)		(mutex_owner(mp) == current)
 #define	MUTEX_HELD(mp)		mutex_owned(mp)
 #define	MUTEX_NOT_HELD(mp)	(!MUTEX_HELD(mp))
--- a/config/spl-build.m4
+++ b/config/spl-build.m4
@@ -49,7 +49,6 @@ AC_DEFUN([SPL_AC_CONFIG_KERNEL], [
 	SPL_AC_USLEEP_RANGE
 	SPL_AC_KMEM_CACHE_ALLOCFLAGS
 	SPL_AC_WAIT_ON_BIT
-	SPL_AC_MUTEX_OWNER
 	SPL_AC_INODE_LOCK
 	SPL_AC_GROUP_INFO_GID
 	SPL_AC_KMEM_CACHE_CREATE_USERCOPY
@@ -1560,35 +1559,6 @@ AC_DEFUN([SPL_AC_WAIT_ON_BIT], [
 	],[
 		AC_MSG_RESULT(no)
 	])
-])
-
-dnl #
-dnl # Check whether mutex has owner with task_struct type.
-dnl #
-dnl # Note that before Linux 3.0, mutex owner is of type thread_info.
-dnl #
-dnl # Note that in Linux 3.18, the condition for owner is changed from
-dnl # defined(CONFIG_DEBUG_MUTEXES) || defined(CONFIG_SMP) to
-dnl # defined(CONFIG_DEBUG_MUTEXES) || defined(CONFIG_MUTEX_SPIN_ON_OWNER)
-dnl #
-AC_DEFUN([SPL_AC_MUTEX_OWNER], [
-	AC_MSG_CHECKING([whether mutex has owner])
-	tmp_flags="$EXTRA_KCFLAGS"
-	EXTRA_KCFLAGS="-Werror"
-	SPL_LINUX_TRY_COMPILE([
-		#include <linux/mutex.h>
-		#include <linux/spinlock.h>
-	],[
-		DEFINE_MUTEX(m);
-		struct task_struct *t __attribute__ ((unused));
-		t = m.owner;
-	],[
-		AC_MSG_RESULT(yes)
-		AC_DEFINE(HAVE_MUTEX_OWNER, 1, [yes])
-	],[
-		AC_MSG_RESULT(no)
-	])
-	EXTRA_KCFLAGS="$tmp_flags"
 ])
 
 dnl #


behlendorf:
Let's get @tuxoko's thoughts on this too.  Personally, I'm happy to shed this bit of cleverness, it's one less way a kernel change can screw things up.

tuxoko:
In #632 
> kernel would use the owner field in mutex in incompatible ways, in particular once a tainted module is loaded (like zfs), the owner field will always be zero triggering mutex_destroy assertion on zfs right away like this

But the owner you get is `ffff88030be28500` not zero. Can you clarify what makes you think the owner field change its meaning?

verygreen:
because if I disable the check for taintedness that disables the lock tracking, then this does not hit right away. You can see it for yourself. Change the EXPORT_SYMBOL_GL to EXPORT_SYMBOL on the mutex_lock_nested(), build zfs agains taht kernel with DEBUG_LOCK_ALLOC enabled and try to run ZFS, it dies right away.
Now commend the check in add_taint():
```
        if (lockdep_ok == LOCKDEP_NOW_UNRELIABLE && __debug_locks_off())
                pr_warn("Disabling lock debugging due to kernel taint\n");
````
to make it never hit and suddenly you can use zfs again most of the time (though it still trips sometimes), now apply this patch and it never hits (but you get issue 6401 occasionally that is a separate use after free on mutexes it appears that I am still digging into).

Additionally, all the other instances of this assertion suddenly disappear which indicates that the kernel uses the field in a way that is incompatible with how spl thinks it's being used.

**Comment by tuxoko** [comment]
But where does it change how the owner field behaves? You do know this assertion you hit is a sanity check, right? If you have other use after free issue, it's perfectly possible you hit this because of some other bug, not because we use the owner field incorrectly.

**Comment by verygreen** [comment]
__debug_locks_off() call changes the behavior so that the assertion hits right away.
So something your code assumes is not true when debug locks is off and once you convince kernel maintainers to export that symbol (as per the comment in the configure check) things would break right away for those who enable the option.
It's probably not true even when that is not called but in subtlier ways or you are racing against kernel usage which is also possible, I guess.

Since this assertion never triggers if you do not reuse in-kernel mutex owner structure, it seems to imply that either your checks with out of mutex code are broken and fail to detect stuff or the code for in-kernel mutex owner are broken and give out false positives (which seems more likely at this point to me).

**Comment by behlendorf** [comment]
@tuxoko any objections to merging this?

**Comment by tuxoko** [comment]
@behlendorf I don't see any point of merging this. There's no sign it had caused any actual problem. And the problem @verygreen had is more likely use-after-free or race problem, or even maybe some subtle memory corruption. And the fact that it's easier to trigger using kernel built-in owner field seems to indicate it's slightly better.

**Comment by verygreen** [comment]
@tuxoco so instant assertion failure is not enough of an actual problem?
The kernel does do different things depending on its internal state and the way it was configured and whatever the spl code does seems to be incompatible at least in some cases. (this is different from the buggy spl use after free problem that I am hittign after applying this patch, mind you, that is fixed by a different patch hinted at by @ironMann in zfs issue 6401 (that I verified is effective)

**Comment by tuxoko** [comment]
> so instant assertion failure is not enough of an actual problem?
It is a problem indeed.
But to fix it we need to figure out what exactly caused it.
If you can provide a way to repro, I can try to figure out the root cause.

**Comment by verygreen** [comment]
Way to repro:
Change your kernel so that mutex_lock_nested is just plain EXPORT_SYMBOL(), not GPL
build with these options:
```
CONFIG_DEBUG_SPINLOCK=y
CONFIG_DEBUG_LOCK_ALLOC=y
CONFIG_LOCKDEP=y
```
build spl/zfs against that kernel
modprobe zfs
attach a block device (i.e. create a a loop device and try to format it) -> assertion failure as described here and in issue 632

Now if you comment out this check in kernel/panic.c::add_taint()
```
        if (lockdep_ok == LOCKDEP_NOW_UNRELIABLE && __debug_locks_off())
                pr_warn("Disabling lock debugging due to kernel taint\n");
```
the crash does not happen right away, but it still does happen at a later point in time

**Comment by tuxoko** [comment]
http://elixir.free-electrons.com/linux/v3.18.63/source/kernel/locking/mutex-debug.c#L83
http://elixir.free-electrons.com/linux/v3.19.8/source/kernel/locking/mutex-debug.c#L89
OK, I see they did change the behaviour of the owner field in older kernel.
This is quite unfortunate, because they fixed this in new kernel.
And this only happens when you are in a weird state of CONFIG_DEBUG_MUTEX but with debug_locks disabled.

**Comment by behlendorf** [comment]
Turns out I should have opened a ZFS PR for this to run it through the buildbot.  Looks like this caused build failures on certain kernel.  I'll be reverting this commit until we can resolve these issues.

http://build.zfsonlinux.org/builders/CentOS%206.7%20x86_64%20%28BUILD%29/builds/15233/steps/shell_2/logs/make

File: include/sys/mutex.h [hunk 1]
```diff
@@ -58,24 +56,16 @@ spl_mutex_set_owner(kmutex_t *mp)
...
 {
-#ifndef HAVE_MUTEX_OWNER
 	mp->m_owner = NULL;
-#endif
 }
 
-#ifdef HAVE_MUTEX_OWNER
 #define	mutex_owner(mp)		(ACCESS_ONCE(MUTEX(mp)->owner))
-#else
-#define	mutex_owner(mp)		(ACCESS_ONCE((mp)->m_owner))
```
tuxoko: Shoud be this one instead of the above one.

verygreen:
oops, I wonder why it did not break anything when it compiled, though?

Ah, I know, because in the actual testing I just undefed the HAVE_MUTEX_OWNER befor creating this proper patch.

include/sys/mutex.h
```diff
@@ -40,10 +40,8 @@ typedef enum {
 typedef struct {
 	struct mutex		m_mutex;
 	spinlock_t		m_lock;	/* used for serializing mutex_exit */
-#ifndef HAVE_MUTEX_OWNER
 	/* only when kernel doesn't have owner */
```
ironMann: remove the comment as well?

behlendorf:
Reverted.  OK, let's take another swing at this.

@verygreen can you open a new PR with an updated patch.  Then we can open a test PR, similar zfsonlinux/zfs#6438, to get a full test run before merging it.  The key bit bit is the ZFS PR with the the `Requires-spl:` line in the commit comment.

verygreen:
Ok, hopefully it works out this time

include/sys/mutex.h
```diff
@@ -58,24 +56,16 @@ spl_mutex_set_owner(kmutex_t *mp)
 	 * kernel will handle its owner, so we don't need to do anything if it
```
ironMann: this comment should also go

behlendorf: I can fix this in the merge.

PR #634, opened 2017-07-25
Module parameter to enable spl_panic() to panic the kernel

verygreen:
In unattended operations it's often more useful to have node
panic and reboot when it encounters problems as opposed to
sit there indefinitely waiting for somebody to discover it.

This implements an spl_panic_crash module parameter, set it
to nonzero to cause spl_panic() to call panic().

Signed-off-by: Oleg Drokin <green@linuxhacker.ru>

### DIFF
--- a/module/spl/spl-err.c
+++ b/module/spl/spl-err.c
@@ -29,6 +29,17 @@
 #include <linux/ratelimit.h>
 
 /*
+ * It is often useful to actually have the panic crash the node so you
+ * can then get notified of the event, get the crashdump for later
+ * analysis and other such goodies.
+ * But we would still default to the current default of not to do that.
+ */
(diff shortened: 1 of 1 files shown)


loli10K:
I've found myself wanting this for a while now, and i've always resorted to patching `spl_panic()` when i needed it (or using systemtap, which is boring in this case).

bunder2015:
Does this cover all instances where zfs/spl would panic, or just the spl ones?  I've come across instances where zfs panicked and the system kept running (sometimes without available pool io) until I hard rebooted.

verygreen:
This covers all cases where spl_panic() is called via various assertions.

ZFS assertions are still done via spl I am sure.

PR #637, opened 2017-08-02
spl-mutex: fix race in mutex_exit

ironMann:
Prevent race on accessing kmutex_t when the mutex is
embedded in a ref counted structure.

Issue zfsonlinux/zfs#6401

Signed-off-by: Gvozden Neskovic <neskovic@gmail.com>

### DIFF
--- a/include/sys/mutex.h
+++ b/include/sys/mutex.h
@@ -183,12 +183,13 @@ spl_mutex_lockdep_on_maybe(kmutex_t *mp)			\
  */
 #define	mutex_exit(mp)						\
 {								\
-	spl_mutex_lockdep_off_maybe(mp);			\
-	spin_lock(&(mp)->m_lock);				\
 	spl_mutex_clear_owner(mp);				\
+	spin_lock(&(mp)->m_lock);				\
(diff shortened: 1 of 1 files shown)


include/sys/mutex.h
```diff
@@ -183,12 +183,12 @@ spl_mutex_lockdep_on_maybe(kmutex_t *mp)			\
...
 #define	mutex_exit(mp)						\
 {								\
-	spl_mutex_lockdep_off_maybe(mp);			\
 	spin_lock(&(mp)->m_lock);				\
+	spl_mutex_lockdep_off_maybe(mp);			\
 	spl_mutex_clear_owner(mp);				\
 	mutex_unlock(MUTEX(mp));				\
-	spin_unlock(&(mp)->m_lock);				\
 	spl_mutex_lockdep_on_maybe(mp);				\
+	spin_unlock(&(mp)->m_lock);				\
```
tuxoko: Can you add a comment here says "no dereference mp beyond this point".

ironMann: Done. 
Additionally, moved `spl_mutex_clear_owner()` outside of the spinlock region.

Issue #638, opened 2017-08-02
Use after free in taskq_seq_show_impl


verygreen:
I have been having a fair amount of crashes on use after free in taskq_seq_show_impl.
Now that the mutex use after free is under control it became the next most frequent crash for me.

The crash typically looks like this:
```
[69476.589393] BUG: unable to handle kernel paging request at ffff8802b300efc8
[69476.607047] IP: [<ffffffffa0979438>] taskq_seq_show_impl+0x1c8/0x400 [spl]
[69476.607754] PGD 2e75067 PUD 33ebfa067 PMD 33ea61067 PTE 80000002b300e060
[69476.608527] Oops: 0000 [#1] SMP DEBUG_PAGEALLOC
[69476.609199] Modules linked in: lustre(OE) ofd(OE) osp(OE) lod(OE) ost(OE) mdt(OE) mdd(OE) mgs(OE) osd_zfs(OE) lquota(OE) lfsck(OE) obdecho(OE) mgc(OE) lov(OE) osc(OE) mdc(OE) lmv(OE) fid(OE) fld(OE) ptlrpc_gss(OE) ptlrpc(OE) obdclass(OE) ksocklnd(OE) lnet(OE) libcfs(OE) zfs(PO) zunicode(PO) zavl(PO) icp(PO) zcommon(PO) znvpair(PO) spl(O) zlib_deflate jbd2 syscopyarea sysfillrect ata_generic sysimgblt pata_acpi ttm drm_kms_helper drm i2c_piix4 ata_piix libata i2c_core serio_raw pcspkr virtio_console virtio_balloon virtio_blk floppy nfsd ip_tables rpcsec_gss_krb5 [last unloaded: libcfs]
[69476.625329] CPU: 2 PID: 12678 Comm: cat Tainted: P           OE  ------------   3.10.0-debug #2
[69476.626650] Hardware name: Red Hat KVM, BIOS 0.5.1 01/01/2011
[69476.627312] task: ffff8802d38d0300 ti: ffff8802d97a4000 task.ti: ffff8802d97a4000
[69476.628527] RIP: 0010:[<ffffffffa0979438>]  [<ffffffffa0979438>] taskq_seq_show_impl+0x1c8/0x400 [spl]
[69476.629810] RSP: 0018:ffff8802d97a7d60  EFLAGS: 00010046
[69476.630451] RAX: ffff8802b300eec0 RBX: ffff8802e8ebee58 RCX: 0000000000000000
[...]

PR #640, opened 2017-08-02
Fix use-after-free in taskq_seq_show_impl

tuxoko:
taskq_seq_show_impl walks the tq_active_list to show the tqent_func and
tqent_arg. However for taskq_dispatch_ent, it's very likely that the
task entry will be freed during the function call, and causes a
use-after-free bug.

To fix this, we duplicate the task entry to an on-stack struct, and
assign it instead to tqt_task. This way, the tq_lock alone will
guarantee its safety.

### DIFF
--- a/module/spl/spl-taskq.c
+++ b/module/spl/spl-taskq.c
@@ -337,19 +337,18 @@ taskq_find_list(taskq_t *tq, struct list_head *lh, taskqid_t id)
 
 /*
  * Find an already dispatched task given the task id regardless of what
- * state it is in.  If a task is still pending or executing it will be
- * returned and 'active' set appropriately.  If the task has already
- * been run then NULL is returned.
+ * state it is in.  If a task is still pending it will be returned.
+ * If a task is executing, then -EBUSY will be returned instead.
+ * If the task has already been run then NULL is returned.
  */
 static taskq_ent_t *
-taskq_find(taskq_t *tq, taskqid_t id, int *active)
+taskq_find(taskq_t *tq, taskqid_t id)
 {
 	taskq_thread_t *tqt;
 	struct list_head *l;
 	taskq_ent_t *t;
 
 	ASSERT(spin_is_locked(&tq->tq_lock));
-	*active = 0;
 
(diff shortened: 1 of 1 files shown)


tuxoko:
https://github.com/zfsonlinux/spl/issues/638

module/spl/spl-taskq.c
```diff
@@ -897,22 +904,24 @@ taskq_thread(void *args)
...
 
 			/*
-			 * In order to support recursively dispatching a
-			 * preallocated taskq_ent_t, tqent_id must be
-			 * stored prior to executing tqent_func.
+			 * A TQENT_FLAG_PREALLOC task may be reused or freed
+			 * during the task function call. Store tqent_id and
+			 * tqent_flags here.
+			 *
+			 * Also use an on disk taskq_ent_t for tqt_task
```
dweeezil: s/on disk/on-stack/

module/spl/spl-taskq.c
```diff
@@ -366,9 +368,13 @@ taskq_find(taskq_t *tq, taskqid_t id, int *active)
 	list_for_each(l, &tq->tq_active_list) {
 		tqt = list_entry(l, taskq_thread_t, tqt_active_list);
 		if (tqt->tqt_id == id) {
-			t = tqt->tqt_task;
 			*active = 1;
-			return (t);
+			/*
+			 * Instead of returning tqt_task, we just return a non
+			 * NULL value to prevent misuse, since tqt_task only
+			 * have two valid fields.
```
dweeezil: s/have/has/

module/spl/spl-taskq.c
```diff
@@ -366,9 +368,13 @@ taskq_find(taskq_t *tq, taskqid_t id, int *active)
...
+			 */
+			return ((taskq_ent_t *)-1L);
```
behlendorf: If we can't safely return the `taskq_ent_t *` in the active case, then we don't we do something like return `ERR_PTR(-EBUSY)` and remove the `active` argument as a simplification.  There are only two callers which would need to be updated.

PR #641, opened 2017-08-08
Allow larger SPA names in stats

gaurkuma:
Signed-off-by: gaurkuma <gauravk.18@gmail.com>

ZFS changes Pull Request: https://github.com/zfsonlinux/zfs/pull/6481

### DIFF
--- a/include/sys/kstat.h
+++ b/include/sys/kstat.h
@@ -32,7 +32,7 @@
 #include <sys/kmem.h>
 #include <sys/mutex.h>
 
-#define KSTAT_STRLEN            31
+#define KSTAT_STRLEN            255
 #define KSTAT_RAW_MAX		(128*1024)
 
 /* For reference valid classes are:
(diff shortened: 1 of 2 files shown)


richardelling:
For the archives: in illumos a method for storing long strings was added in commit 0e96d867a045da50bf7e014cab94cb44528a0f5 in July 2015. The method is described thusly,
```
 *      KSTAT_FLAG_LONGSTRINGS:
 *
 *              Indicates that this kstat contains long strings (which
 *              are stored outside of the kstat data section). When copied
 *              out to user space the string data will be held in the data
 *              section provided by the user.
```

As you can imagine, this is a PITA to use because you have to manage the string's location in-kernel. Specifically, remember when changing it's allocation to also update the kstat's pointer. Also, for SPL specifically, the user code would need to be ported. 

The solution implemented here is much simpler and less error-prone for future developers.

For some historical perspective, there was a time when Solaris ran on a workstation with 4MB of RAM. Every byte was precious. Thankfully, those days are long gone.

PR #645, opened 2017-08-12
Test only - Prevent SPL Build for testing  [WIP]

ofaaland:
For use testing merge of SPL into ZFS, https://github.com/zfsonlinux/zfs/pull/6479, via buildbot.


ofaaland:
Will accomplish this through buildbot since we need to support build builds with a separate SPL for the stable branches even after the SPL merge patch is landed on ZFS master.

PR #646, opened 2017-08-23
Multi-buffer sha256 support in SPL to ZFS

dong-liuliu:
This patch implements multi-buffer hash facilities to export a general
SHA256 api to ZFS. When multibuffer hash is configured in SPL, its
SHA256 api for ZFS will automaticlly transfer SHA256 compute task to
multi-buffer hash facilities which will increase SHA256 performance
2~7 times.

ISA-L multi-buffer hash provides software acceleration for SHA256:
https://github.com/01org/isa-l_crypto
* To enable multi-buffer hash acceleration, you need to download ISA-L
crypto lib source code from above link and checkout master branch.
* Configure multi-buffer hash in SPL, e.g.:
 ./configure --with-hash-mb=<isa-l-crypto-code-path>
 make
  Following make notice, run:
 make -C <isa-l-crypto-code-path> -f Makefile.kern kernobj
 Then, continue run:
 make
* After that, a C predefine marco HAVE_HASH_MB and C api mulbuf_sha256
will be exported out to ZFS module

If ZFS is also patched with multi-buffer sha256, then.
Set sha256 checksum in ZFS dataset: "checksum = sha256"
Data written to ZFS pool will be checksumed by multi-buffer sha256.
Run write workload, on Intel Xeon of Haswell type, cpu consumption of
sha256 will be decreased to be previous 1/2~1/5.

Multibuffer sha256 follows standard sha256 format. With SIMD instruction
set, it can process 4/8/16 sha256 tasks simutaneously. So it has a much
better compute throughput.
Here are Isa-L's performance data sheets about multi-buffer hash.

Intel®   Xeon® E5-2600v3 (Haswell) |   |  
-- | -- | --
Hash functions | ISA-L multi-buffer version | OpenSSL single version
SHA-1 | 1.9 GB/s | 542 MB/s
SHA-256 | 881 MB/s | 183 MB/s
SHA-512 | 689 MB/s | 286 MB/s
MD5 | 3.7 GB/s | 465 MB/s

Intel® Xeon® Platinum 8180 (Skylake) |   |  
-- | -- | --
Hash functions | ISA-L multi-buffer version | OpenSSL single version
SHA-1 | 5.4 GB/s | 605 MB/s |  
SHA-256 | 2.7 GB/s | 216 MB/s |  
SHA-512 | 2.2 GB/s | 334 MB/s |  
MD5 | 9.7 GB/s | 401 MB/s |

### DIFF
--- a/config/spl-build.m4
+++ b/config/spl-build.m4
@@ -17,6 +17,7 @@ AC_DEFUN([SPL_AC_CONFIG_KERNEL], [
 	KERNELCPPFLAGS="$KERNELCPPFLAGS -Wstrict-prototypes"
 	AC_SUBST(KERNELCPPFLAGS)
 
+	SPL_AC_HASH_MB
 	SPL_AC_DEBUG
 	SPL_AC_DEBUG_KMEM
 	SPL_AC_DEBUG_KMEM_TRACKING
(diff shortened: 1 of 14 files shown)


tonyhutter:
Does the performance stay the same if you're using an older-than-Haswell CPU?

dong-liuliu:
On Xeon Sandy Bridge or Ivy Bridge, it is still better. On others, it should be the same.

behlendorf:
@dong-liuliu sorry about the delay but the spl has now been merged to the zfs repository. If you're still interested in getting this support included please go ahead and open a new PR against ZFS.

thegreatgazoo:
I thought isa-l_crypto was a user-space library. If I was not wrong, could you please describe how SPL could use that library from the kernel?

dong-liuliu:
Yes,  originally, it's a user-space library. Since it only do computing without further logical operations, it is easy to port isa-l_crypto into kernel env with a wrapper layer.

thegreatgazoo:
@dong-liuliu is there an official kernel port already? Thanks!

PR #654, opened 2017-10-04
Fix build on kernel 4.13 onwards

fahadsadah:
Build fails on kernels 4.13 onwards, which enable struct randomisation (https://github.com/torvalds/linux/commit/3859a271a003aba01e45b85c9d8b355eb7bf25f9) invalidating positional struct accesses

The spl-proc.c file has a couple of positional struct accesses, removed in v0.7. This PR cherry-picks commit 120faefed90aa4c276a3db45525dc5c804cb1a00 into the spl-0.6.5-release branch to remove them here too and fix the build :)

### DIFF
--- a/module/spl/spl-proc.c
+++ b/module/spl/spl-proc.c
@@ -402,11 +402,11 @@ static struct ctl_table spl_kmem_table[] = {
                 .mode     = 0444,
                 .proc_handler = &proc_doslab,
         },
-	{0},
+	{},
 };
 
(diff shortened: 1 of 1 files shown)


behlendorf:
@fahadsadah thanks for identifying and porting this fix.  However, since we do not plan on tagging any new 0.6.5.x releases, and the fix is already in the 0.7.x series, I'm going to close this PR.

PR #658, opened 2017-10-07
Fix reaping Linux kmem caches

behlendorf:
When the kmem cache is backed by a Linux slab cache requests
to reap from the cache should be passed along.

Add missing parenthisis around btop and ptob macros to ensure
operation ordering is preserved after expandsion.

### DIFF
--- a/include/sys/param.h
+++ b/include/sys/param.h
@@ -28,8 +28,8 @@
 #include <asm/page.h>
 
 /* Pages to bytes and back */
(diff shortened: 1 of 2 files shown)


tuxoko:
We've been there before, but kmem_cache_shrink is not used for this purpose.

https://github.com/zfsonlinux/spl/pull/487

behlendorf:
Ahh yes.  Thank you, I'd already forgotten about that.  In that case, let me close this and simply fix the missing parenthesis.

PR #659, opened 2017-10-09
    Linux 4.14 compat: vfs_read/write

rfehren:
Commit torvalds/linux@bd8df82b
    - Unexport vfs_read and vfs_write

Patch received basic run-time testing applied to the spl-0.7-release branch and built against 4.14-rc3.

### DIFF
--- a/module/spl/spl-vnode.c
+++ b/module/spl/spl-vnode.c
@@ -235,9 +235,17 @@ vn_rdwr(uio_rw_t uio, vnode_t *vp, void *addr, ssize_t len, offset_t off,
         set_fs(get_ds());
 
 	if (uio & UIO_WRITE)
+#ifdef HAVE_VFS_WRITE
 		rc = vfs_write(fp, addr, len, &offset);
+#else
+		rc = kernel_write(fp, addr, len, &offset);
+#endif /* HAVE_VFS_WRITE */
 	else
+#ifdef HAVE_VFS_WRITE
 		rc = vfs_read(fp, addr, len, &offset);
+#else
+		rc = kernel_read(fp, addr, len, &offset);
+#endif /* HAVE_VFS_WRITE */
 
 	set_fs(saved_fs);
 	fp->f_pos = offset;
--- a/config/spl-build.m4
+++ b/config/spl-build.m4
@@ -1,4 +1,4 @@
-###############################################################################
+##############################################################################
 # Copyright (C) 2007-2010 Lawrence Livermore National Security, LLC.
 # Copyright (C) 2007 The Regents of the University of California.
 # Written by Brian Behlendorf <behlendorf1@llnl.gov>.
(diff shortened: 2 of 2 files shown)


module/spl/spl-vnode.c
```diff
@@ -235,9 +235,17 @@ vn_rdwr(uio_rw_t uio, vnode_t *vp, void *addr, ssize_t len, offset_t off,
...
 		rc = vfs_write(fp, addr, len, &offset);
+#else
+		rc = kernel_write(fp, addr, len, &offset);
+#endif /* HAVE_VFS_WRITE */
 	else
+#ifdef HAVE_VFS_WRITE
 		rc = vfs_read(fp, addr, len, &offset);
+#else
+		rc = kernel_read(fp, addr, len, &offset);
+#endif /* HAVE_VFS_WRITE */
```
behlendorf: It looks like `kernel_write()` was exported in v3.8 torvalds/linux@7bb307e894d51308aa0582a8c4cc5875bbc645b9 and `kernel_read()` in 3.9 torvalds/linux@3dc20cb282ec03cc4c997130d680c800011ed479.  This is arguably when we should have moved over to these new interfaces.

What we need to do handle this is to add a configure check for both `kernel_read` and `kernel_write` and a compatibility wrapper for each of these called `spl_kernel_read()` and `spl_kernel_write()` for consistency.  The wrappers should prefer the `kernel_*` interfaces if they're determined to be available.  And since these functions are only called from `spl-vnode.c` they can be added there and declared static.

There are a couple things to be careful of when implementing the wrappers.

* The arguments for the functions do not appear in the same order.
* `vfs_read/write` functions don't handle the `set_fs()` bit but the `kernel_read/write` functions do.  This logic will need to be pulled in to the wrapper for only the `vfs_read/write` case and out of `vn_rdwr`.

rfehren: Unfortunately, I won't have the time to work further on this issue. Concerning your comments: I would actually leave it the way it is. All previous ZFS versions compiled against kernels < 4.14 have been using vfs_read/write and were tested against this. Why risk catching a regression with such kernels due to a possibly buggy early implemenation of kernel_read/write? I don't see a benefit, only potential problems when using kernel_read/write with kernels < 4.14.
I signed off the commit rebased and pushed.

bmaxa:
I have problem with this patch, cannot mount files on older kernels. Drives mount ok.
I also have this errors in dmesg on kernel 4.14 when importing pools,
 noop error is because I use blk-mq and there noop is called none.
  143.026192] attempt to access beyond end of device
[  143.026194] unknown-block(8,0): rw=0, want=468862135, limit=468862128
[  143.026195] Buffer I/O error on dev sda1, logical block 58607758, async page read
[  144.193110] attempt to access beyond end of device
[  144.193115] unknown-block(8,16): rw=0, want=3907029175, limit=3907029168
[  144.193119] attempt to access beyond end of device
[  144.193121] unknown-block(8,16): rw=0, want=3907029175, limit=3907029168
[  144.193123] Buffer I/O error on dev sdb1, logical block 488378638, async page read
[  145.009731] ZFS: Unable to set "noop" scheduler for /dev/disk/by-id/ata-ST2000VN000-1H3164_W1H2DSKA-part1 (sdc): 256
[  145.256794] ZFS: Unable to set "noop" scheduler for /dev/disk/by-id/ata-ST2000VN000-1H3164_W1H2DSKA-part1 (sdc): 256

behlendorf:
Replaced by #667.

PR #674, opened 2017-12-28
Add cv_timedwait_io()

behlendorf:
Add missing helper function cv_timedwait_io() which should be used
when waiting on IO with a specified timeout.


behlendorf:
It looks like `io_schedule_timeout()` wasn't exported in the 2.6.32 kernel so we'll need to figure out a reasonable way to handle that before this can be merged.

behlendorf:
I've addressed the 2.6.32 issue by implementing our own `io_schedule_timeout()` by registering a timer to wake the process and calling `io_schedule()`.

PR #678, opened 2018-01-17
Fix Debian packaging on ARMv7/ARM64

loli10K:
When building packages on Debian-based systems specify the target architecture used by `alien` to convert .rpm packages into .deb: this avoids detecting an incorrect value which results in the following errors:

```
<package>.aarch64.rpm is for architecture aarch64 ; the package cannot be built on this system
<package>.armv7l.rpm is for architecture armel ; the package cannot be built on this system
```

This change is intended to fix https://github.com/zfsonlinux/zfs/issues/7046.


gmelikov:
@Menion2k ZFS is already patched too https://github.com/zfsonlinux/zfs/pull/7058

Issue #684, opened 2018-02-17
VERIFY*() macros evalulate parameters multiple times on failure


DeHackEd:
Someone in IRC just had a crash in their system with the following error: https://paste.pound-python.org/show/gjjdCxPcRYtmVhyqVxXu/ What's odd is that the statement `(0 == 0)` implies the result should have been a success. In fact the VERIFY3U macro double-evaluated the `arc_buf_alloc_impl` call which succeeded the second time. That's a bug.

https://github.com/zfsonlinux/spl/blob/master/include/sys/debug.h#L66

PR #685, opened 2018-02-17
Fix multiple evaluations of VERIFY() and ASSERT() on failures

DeHackEd:
Signed-off-by: DHE <git@dehacked.net>


DeHackEd:
Added a reference to #684 so github will auto-close it.

I should have just open the pull request and not bothered with the issue tracker.

PR #686, opened 2018-02-19
PR for trivial SPL fixes.

kusumi:
PR for trivial SPL fixes.


behlendorf:
4bc3286 ￼Staticize kstat_default_update() ￼…
3157148 Fix function name typos

These two look good, but I'd like to hold off on `kobj_open_file()` cleanup for the moment.  The truth is the `kobj` interface is very lightly used, there are only two callers I believe, and I'd much rather retire it entirely if possible in the near future as a similification.

PR #704, opened 2018-05-06
Update README: run autogen first

ahrens:


README.markdown
```diff
@@ -8,6 +8,7 @@ primitives.
 
 To build packages for your distribution:
 
+    $ sh autogen
```
dinatale2: Shouldn't this be `sh autogen.sh`?

ahrens: yes!  will fix, thanks.

ahrens:
Now I realize that the README does cover this in the next paragraph: `If you are building directly from the git tree and not an officially released tarball you will need to generate the configure script. This can be done by executing the autogen.sh script after installing the GNU autotools for your distribution.`

It seems like the instructions are coming from the point of view of building packages from an "officially released tarball".  I wonder if that should be the primary audience here, vs. folks that want to compile SPL from the source on github (so that they can compile ZFS from the source on github).  For example, `make pkg` is insufficient to build what zfs needs (`Module.symvers`, for example).  Instead you would want to run `make`.

behlendorf:
Yes, the expected audience was primarily users building from the tarballs.  However, enough people have been tripped up by this exact thing that I think it's reasonable to list it explicitly.  It's not harmful to those building from released tarballs and it does avoid one potential problem.

> make pkg

This should be sufficient, but you do need to install the resulting development packages.  That unfortunately isn't clearly explained in the README.  We are working to updated wiki documentation to make this more clear.

PR #705, opened 2018-05-25
Add KSTAT_FLAG_NO_HEADERS for kstats

tonyhutter:
The normal kstat output includes two metadata headers before the actual value:
```
    $ cat /proc/spl/kstat/zfs/foo
    31 0 0x01 1 0 1336141223666 1816471471871
    raw value
    foo
```
With this patch, you can omit those headers by setting `KSTAT_FLAG_NO_HEADERS`:
```
    $ cat /proc/spl/kstat/zfs/foo
    foo
```
This will be used by an upcoming ZFS patch that I'm working on.


gmelikov (approved):

tonyhutter:
Now that SPL is merged, I folded these changes into https://github.com/zfsonlinux/zfs/pull/7563

PR #711, opened 2018-12-05
Use autoconf variable for C preprocessor

lopsided98:
This fixes the same issue as in https://github.com/zfsonlinux/zfs/pull/8180, except in the SPL. This fixes detection of the kernel version when cross-compiling without access to a native compiler.


tonyhutter (approved):

tonyhutter:
Thanks, I'll pull this in on the next release.

gmelikov (approved):

PR #716, opened 2019-05-15
Linux 5.1 compat: Fix get_ds()

gamanakis:
get_ds() is no longer provided in the 5.1 kernel.
Define get_ds() in that case.

Signed-off-by: George Amanakis <gamanakis@gmail.com>

### DIFF
--- a/module/spl/spl-vnode.c
+++ b/module/spl/spl-vnode.c
@@ -29,6 +29,11 @@
 #include <sys/kmem_cache.h>
 #include <linux/falloc.h>
 #include <linux/file_compat.h>
+#include <linux/version.h>
+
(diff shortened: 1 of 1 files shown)


tonyhutter:
Thanks for this!

Note that we typically detect the presence of deprecated kernel functions via autotools scripts rather than doing the explicit kernel version checks.  See https://github.com/zfsonlinux/zfs/commit/bbb1b6cea75ff8ff27742ce778664ddeec371a01 for an example.  You'll also want to base your code off the master branch rather than spl-0.7-release.

gamanakis:
Actually this can be resolved much more easily, see:
https://github.com/zfsonlinux/zfs/commit/b46fd243d57d0af4b4b568d8c0c4b30b79059ecc

